PR #123, opened 2017-06-20
Frontend dev

mjoyce91:


.codeclimate.yml
```diff
@@ -1,3 +1,7 @@
-## ignore js dependencies
+JavaScript: truu
```
jseppi: `true`?

jseppi:
You might want to disable the `import/extensions` check in CodeClimate's JS engine.
Here's an example `.codeclimate.yml` from an 18F project:

```yml
engines:
  eslint:
    enabled: true
    channel: "eslint-3"
    checks:
      import/no-unresolved:
        enabled: false
      import/extensions:
        enabled: false
    extensions:
    - .js
    - .jsx
```

public/assets/uswds/CONTRIBUTING.md
```diff
@@ -0,0 +1,145 @@
+## Welcome!
```
burgwyn: Let's delete this

public/index.html
```diff
@@ -0,0 +1,174 @@
...
+    <!--
+      Notice the use of %PUBLIC_URL% in the tags above.
+      It will be replaced with the URL of the `public` folder during the build.
+      Only files inside the `public` folder can be referenced from the HTML.
+
+      Unlike "/favicon.ico" or "favicon.ico", "%PUBLIC_URL%/favicon.ico" will
+      work correctly both with client-side routing and a non-root public URL.
+      Learn how to configure a non-root public URL by running `npm run build`.
+    -->
+    <title>React App</title>
```
burgwyn: TalentMAP?

public/manifest.json
```diff
@@ -0,0 +1,15 @@
+{
```
burgwyn: Let's change some of these values - names and `theme_color`

src/Components/App/App.css
```diff
@@ -0,0 +1,24 @@
+.App {
+  text-align: center;
+}
+
+.App-logo {
```
burgwyn: Are we still using the App logo?

src/Components/App/App.css
```diff
@@ -0,0 +1,24 @@
...
+  height: 150px;
+  padding: 20px;
+  color: white;
+}
+
+.App-intro {
+  font-size: large;
+}
+
+@keyframes App-logo-spin {
```
burgwyn: Remove this animation

src/Components/Results/Results.js
```diff
@@ -0,0 +1,59 @@
...
+    super(props);
+    this.state = {
+      posts: [],
+    };
+  }
+
+  componentDidMount() {
+    const script = document.createElement('script');
+
+    script.src = '/assets/uswds/js/uswds.min.js';
```
burgwyn: Should we pull this out and move load in `index.html`?

src/logo.svg
```diff
@@ -0,0 +1,7 @@
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 841.9 595.3">
```
burgwyn: Delete

src/registerServiceWorker.js
```diff
@@ -0,0 +1,51 @@
+// In production, we register a service worker to serve assets from local cache.
```
burgwyn: We can leave the service worker in for now...

jseppi:
What about bringing in the USWDS some way other than copy-pasting the files into this repo? It's available on npm, so you can add it to your `package.json`, which then opens up some other options for bringing it into the project:

1) My favorite: Use webpack with `ExtractTextPlugin`, `file-loader`, `css-loader`, and `sass-loader`. With this approach you can just do something like `@import '../../node_modules/uswds/dist/css/uswds.css';` in your main SASS file.
Here's an example PR where I recently used this approach: https://github.com/18F/federalist/pull/966/files

2) Use an npm script to copy the files into the local repo, like on `postinstall`. You'd add whatever the local destination path is to your `.gitignore` so the USWDS files don't actually get committed to the repo, but they'll be available locally to the project.

mjoyce91:
Thanks @jseppi - looks like the create-react-app uses `url-loader` instead of `file-loader` for its ability to embed smaller assets. Any thoughts on using one over the other?

jseppi:
Hrm, `url-loader` might work pretty well since I think all the assets in USWDS are fairly small. Good idea!

mjoyce91:
Code coverage >90% and no issues returned from code climate.

PR #127, opened 2017-06-23
Adding scss-lint to CodeClimate environment

burgwyn:
- Adding configuration
- Linting fixes

### DIFF
--- a/src/sass/_home.scss
+++ b/src/sass/_home.scss
@@ -1,3 +1,3 @@
-.Home {
-
+.home {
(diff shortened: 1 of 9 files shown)


src/sass/_home.scss
```diff
@@ -1,3 +1,3 @@
-.Home {
-
+.home {
```
mjoyce91: While there are no associated styles with it, the `Home` component references the previously uppercase `.Home` class.

PR #175, opened 2017-06-30
Update .pa11y-ci config file structure for correct usage of standard,…

mjoyce91:
… apply wait to allow pages to load

Two problems with previous `.pa11y-ci` config file.
- `standard` needed to be wrapped in a `default` object
- `wait` needs to be applied to allow ajax requests to complete

### DIFF
--- a/.pa11yci
+++ b/.pa11yci
@@ -6,6 +6,6 @@
     "urls": [
         "http://localhost:3000",
         "http://localhost:3000/#/results",
-        "http://localhost:3000/#/details/1"
+        "http://localhost:3000/#/details/00025003"
     ]
 }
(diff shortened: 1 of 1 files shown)


.pa11yci
```diff
@@ -1,8 +1,11 @@
 {
+    "defaults": {
+        "wait": 2000,
+        "standard": "WCAG2AA"
+    },
     "urls": [
         "http://localhost:3000",
         "http://localhost:3000/#/results",
         "http://localhost:3000/#/details/1"
```
burgwyn: Does it matter here on the path that we're expecting an object with `id = 1`?

mjoyce91: Good point, this is an old from the mock API I was using. I will update to a real one.

Issue #122, opened 2017-06-19
View Post Profile - Mission Data


burgwyn:
As a bidding employee,
I would like view a post profile's mission data
So I can research on a post

**Acceptance Criteria**

* Display specific post mission data
  * Country
  * City
  * Type of post - eg, embassy, consulate
  * Inter and intra agency representation
    * Points of Contact
  * Language(s)
  * R & R alignment
  * Danger pay
  * Post differential pay
  * Service needs differential
  * COLA
  * Consumables

mjoyce91:
Currently using placeholder data for **type** (hard coded in HTML as "Embassy") and **Inter and intra agency representation - Points of Contact** (hard coded in HTML as "John Doe") on Post profiles.

Issue #30, opened 2017-06-07
Position description


wongperson:
As a bidding employee,
I want to be able to see expanded details about the position
So I can further understand whether the position matches my professional preferences

**Acceptance criteria:**

- Upon selecting the position from the search results, the bidder should be able to see the written description and requirements of the position
- The bidder should be able to see post and position information, including:
    - Skill code
    - Danger pay
    - Language requirements
    - Position grade
    - Region
    - Post
    - Post differential
    

**Definition of done:**
Assumptions documented
Coded work is complete
Full unit test coverage
Product owner approval

apbdevine:
Edited acceptance criteria to remove Tour of Duty length based on information received from DOS team. "the standard TOD is associated with a post. We do not maintain a TOD for a position. When assignments are initiated the TOD defaults to the Post's. However, the assignment TOD can be changed by selecting from the common list of TODs or creating a customized TOD."

mjoyce91:
Using placeholder data for **Region**.

Issue #28, opened 2017-06-07
Search results


wongperson:
As a bidding employee,
I want to see the positions that match all of my search filter criteria
So I can select a potential position to examine further

**Acceptance criteria:**
- The bidder should be able to see all of the position titles available to them based on their search parameters
- The bidder should be able to understand which post the position is located
- The bidder should be able to understand if there is differential pay associated with the position

**Definition of done:**
Assumptions documented
Coded work is complete
Full unit test coverage
Product owner approval

PR #192, opened 2017-07-07
Post profiles

mjoyce91:
This PR adds a new view, Post Profiles, which is integrated with the API. Post data is displayed in other areas, such as Position Details and Results.

#122 - Post profile pages integrated
#30 - Show all fields listed
#28 - Show all fields listed

Real data from the API is used when possible, but the following are mocked with static data:
- **Region** on Position Description
- **Type** on Post Profile
- **Inter and intra agency representation - Points of Contact** on Post Profile


src/Components/PositionDetails/PositionDetails.jsx
```diff
@@ -15,7 +16,7 @@ class ResultsList extends Component {
       )) : <span key="no-languages"> None listed </span>;
     return (
       <div className="usa-grid-full">
-        <div style={{ backgroundColor: '#DFDFDF', marginTop: '10px', marginBottom: '10px', padding: '15px 30px' }}>
+        <div style={{ backgroundColor: '#F2F2F2', marginTop: '10px', marginBottom: '10px', padding: '15px 30px' }}>
```
burgwyn: 🎨

src/Components/PostMissionData/PostMissionData.jsx
```diff
@@ -0,0 +1,77 @@
...
+      <div className="usa-grid-full">
+        <div>
+          <p>
+            Location: {post.description}
+            <br />
+            Tour of duty: {post.tour_of_duty}
+            <br />
+            Type: Embassy
+            <br />
+            POC: John Doe
```
burgwyn: Add a comment for this and any other hard-coded field - // TODO - hard code, etc

src/Components/PositionDetails/PositionDetails.jsx
mjoyce91: Links needed more contrast against the background to pass accessibility tests :) 🎨

PR #195, opened 2017-07-10
Moving API to environment variable

burgwyn:
- Adding env variable
- Changing app binding


burgwyn:
Closing this pull request and adding this behavior to the Node.js app

Issue #118, opened 2017-06-19
Favorite Position


burgwyn:
As a bidding employee,
I want to be able to mark a position
So I can save it for future research

**Acceptance Criteria**

* Users can mark position for other purposes, such as comparison
* User sees a visual confirmation that the mark has been saved
* Marking a position is persisted across browsing sessions

PR #197, opened 2017-07-10
Sprint 2

mjoyce91:
This covers the work completed in Sprint 2

- #28 - Search Results updates
- #30 - Position Description updates
- #118 - Position Favoriting
- #122 - Post Profile Mission Data


src/Containers/Home/Home.jsx
```diff
@@ -143,17 +152,17 @@ class Home extends Component {
...
-                    value={selection.position_number__icontains}
+                    value={searchText.value}
                     onChange={e => this.changeText(e)}
                     type="search"
                     name="search"
                   />
                   <div id="enabled-search" className={enableSearch}>
-                    <a href={`/#/results?${qString}`}>
+                    <Link to={`/results?${qString}`}>
                       <button type="submit">
```
jseppi: Wrapping a `<button>` in a `<Link>`/`<a>` seems a bit odd to me. I think using a `<form>` as mentioned above would help simplify this. You'd have to do the routing in an `onSubmit` handler, like in the example at https://facebook.github.io/react/docs/forms.html#controlled-components

jseppi: And you could use the `disabled` attribute in the submit button instead of the nearly duplicate `<div>`s (ie, the ones with `id="enabled-search" className={enableSearch}`/`id="disabled-search" className={disableSearch}`).

src/Containers/Post/Post.jsx
```diff
@@ -0,0 +1,51 @@
...
+class Post extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      post: {},
+    };
+  }
+
+  componentWillMount() {
+    this.getPost(this.context.router.route.match.params.id);
```
jseppi: This very long string of property accessing makes me a bit nervous. I _think_ once this project is switched to `Redux` you'll probably want to use [`react-router-redux`](https://github.com/reactjs/react-router-redux), which will hopefully help avoid things like this. 

https://github.com/reactjs/react-router-redux#how-do-i-access-router-state-in-a-container-component

src/Containers/Results/Results.jsx
```diff
@@ -0,0 +1,50 @@
...
+  }
+
+  render() {
+    const { results } = this.state;
+    return (
+      <div>
+        {results.length ?
+          <ResultsList results={results} />
+          :
+          <div className="usa-grid-full" />
```
jseppi: A simple message shown to the user in the case of no results would probably be good :)

src/Components/FavoritesButton/FavoritesButton.jsx
```diff
@@ -0,0 +1,87 @@
...
+        </button>
+      </div>
+    );
+  }
+}
+
+FavoritesButton.propTypes = {
+  refKey: PropTypes.string.isRequired,
+  type: PropTypes.string.isRequired,
+  iterator: PropTypes.number,
```
mjoyce91: @jseppi - I was using a generic `id` for the `<button>` tags, so this allows a number to be passed in to provide a unique ID (such as when the results list is being rendered).

src/Containers/Results/Results.jsx
mjoyce91: @jseppi - in progress in our redux branch! 👍

src/Components/FavoritesButton/FavoritesButton.jsx
```diff
@@ -0,0 +1,87 @@
...
+      localStorage.setItem(this.state.localStorageKey,
+        JSON.stringify(existingArray));
+    }
+    this.setState({ saved: !this.state.saved });
+  }
+
+  render() {
+    let fromText = null;
+    fromText = this.props.type === 'fav' ? 'Favorites' : fromText;
+    fromText = this.props.type === 'compare' ? 'Comparison' : fromText;
```
mjoyce91: @jseppi Separate components sound like a good idea, especially if the logic for "favoriting" vs "comparing" starts to change as the app matures (different styles, saving to the user's profile vs the session, etc.)

src/Containers/Home/Home.jsx
mjoyce91: @jseppi - as I'm going back to review this, I'm remembering why I set it up this way. I wanted to use the default look and feel of the USWDS search bar (with the input and submit button connected to each other), which requires `type="search"` on the `<input>`. However, when that's within a form, there's some javascript attached to it which pushes `?search=<your search phrase>` to history, which is not what we want, because then we end up with `.../?search=german` instead of `.../results?q=german&<all our other filters>`.

src/Components/FavoritesButton/FavoritesButton.jsx
mjoyce91: Based on your other comment on using `wrapper.find('button')`, this may not be necessary any more.

src/Components/FavoritesButton/FavoritesButton.jsx
```diff
@@ -0,0 +1,87 @@
...
+    };
+  }
+
+  componentWillMount() {
+    const localStorageKey = this.props.type;
+    this.setState({ localStorageKey });
+  }
+
+  componentDidMount() {
+    const retrievedKey = localStorage
```
mjoyce91: Probably another good piece of functionality to move to redux

src/Containers/Home/Home.jsx
jseppi: I think in your onSubmit handler you could call `e.preventDefault()` first.

Here's some quick untested code to explain what I mean:

in `render()`:

```jsx
<form onSubmit={e => this.onSubmit(e)}>
  <label className="usa-sr-only" htmlFor="search-field">
      Search
  </label>
  <input
    id="search-field"
    value={searchText.value}
    onChange={e => this.changeText(e)}
    type="search"
    name="search"
  />
  <div id="enabled-search" className={enableSearch}>
    <button type="submit" disabled={this.shouldDisableSearch()}>
      <span className="usa-search-submit-text">Search</span>
    </button>
  </div>
</form>
```

and then `onSubmit` would look something like:

```js
onSubmit(e) {
  e.preventDefault();
  // use router API to route to `/results?${qString}`
}
```

mjoyce91: That seemed to work with the inclusion of `withRouter`, but might be better left for the redux integration.

jseppi:
Want to go ahead and merge this so that we're ready for the sprint 3 code review on Monday?

PR #208, opened 2017-07-11
Clean up favorites/compare button text, export PropTypes and some tex…

mjoyce91:
…t as constants,

Addresses feedback from #197 , excluding some items that will be part of the redux integration.


src/Constants/AlertMessages.js
```diff
@@ -0,0 +1,4 @@
+export const NO_LANGUAGES = 'None listed';
```
burgwyn: 👍

PR #212, opened 2017-07-19
Redux

mjoyce91:
This PR includes changes to introduce Redux into the application. Connected components (using `connect` from react-redux) are used in order to provide presentational components with data. Reducers and actions were created to retrieve filters, position details, posts, and search results.


src/Components/PostDetails/PostDetails.jsx
```diff
@@ -0,0 +1,65 @@
...
+    );
+    return (
+      <div>
+        {postMissionData}
+      </div>
+    );
+  }
+}
+
+PostDetails.propTypes = {
```
burgwyn: Do we want to externalize these proptypes?

src/Constants/PropTypes.js
```diff
@@ -25,12 +25,12 @@ export const POST_MISSION_DATA = PropTypes.shape({
 });
 
 export const DETAILS = PropTypes.shape({
```
burgwyn: Let's rename to a more specific kind of `DETAILS`

src/Containers/Home/Home.test.jsx
```diff
@@ -1,128 +1,24 @@
...
-          { id: 3, code: 'AB', long_description: 'Albanian', short_description: 'Albanian', effective_date: '1984-05-04' },
-          { id: 4, code: 'AC', long_description: 'Amharic', short_description: 'Amharic', effective_date: '1984-05-04' },
-    ],
-    [
-          { id: 2, code: '00' },
-          { id: 3, code: '01' },
-    ],
-  ];
+describe('Main', () => {
+  const api = 'http://localhost:8000/api/v1';
```
burgwyn: Is it possible to change the host name/port?  Can we use something that references this is a unit test?

src/Components/PostDetails/PostDetails.jsx
mjoyce91: Good catch - missed this one.

mjoyce91: Resolved with 47b4bac

src/Constants/PropTypes.js
mjoyce91: Updated with `POSITION_DETAILS`

src/Containers/Home/Home.test.jsx
burgwyn: That's an idea.  The scenario I'm thinking of is `ctl+F` on 'localhost' and seeing test files in the mix.  We can table this for later.

Issue #117, opened 2017-06-19
Compare Positions


burgwyn:
As a bidding employee,
I want to be able to compare the details about two positions
So I can further understand whether a position matches my professional preferences relative to another position

**Acceptance Criteria**

* Initiated via search results based on a user marking or selection
* Displays two positions side-by-side
* Displays a list fields from a Position
  * Position Identifier
  * Title
  * Post Name
  * Region
  * Bureau 
  * Tour of duty
  * Differential
  * Danger pay
  * COLA
  * Language(s)
  * Grade

Issue #29, opened 2017-06-07
Clear search filters


wongperson:
As a bidding employee,
I want to set all of my search filter criteria to their default selection
So I can start my search process over with a new combination of search criteria

**Acceptance criteria:**
- The bidder should no longer see search results
- The bidder should be returned to the search filters/parameters default view
- The bidder should see that all selections for their search filters are reset to their default selections
- The bidder should receive a prompt before clearing search filters

**Definition of done:**
Assumptions documented
Coded work is complete
Full unit test coverage
Product owner approval

burgwyn:
Closing

PR #221, opened 2017-07-20
Compare positions, reset filters

mjoyce91:
Adds functionality from #117 :
- Adds functionality to compare positions by making use of the existing "Add/remove from comparison" button using local storage.
- When at least one position exists within the local storage comparison array, a new link and button appear in the Results page - one a link to Compare Positions which shows a side by side of the positions; and another a button to reset any selected comparisons to an empty array.
- New `/compare:ids` route that expects a list of positions numbers, such as `/compare/0005BA6,1001CM7`.
- Additionally, local storage `gets`/`sets` were externalized to `utilities.js` for use across multiple components and abstraction.

Adds functionality from #29 :
- Because of current site structure, this simply returns the user to the home page. By default, filters are not saved, so returning the home page implicitly meets the requirement for the search filters to be reset.
- The user must click the button twice, as the requirements state the user must receive a prompt to confirm that they want to clear their search filters. The button changes text from "Reset Filters" to "Are you sure", which upon clicking returns the user to the home page.


src/Components/CompareList/CompareList.jsx
```diff
@@ -0,0 +1,71 @@
...
+class CompareList extends Component {
+
+  onChildToggle() {
+    this.forceUpdate();
+    this.props.onToggle();
+  }
+
+  render() {
+    const { compare } = this.props;
+    // TODO - move language list to its own component, as this is used in multiple places
```
burgwyn: Can you make an issue for this?

src/Components/ResetFilters/ResetFilters/ResetFilters.jsx
```diff
@@ -0,0 +1,39 @@
+import React, { Component } from 'react';
```
burgwyn: This looks double-nested in the folder

burgwyn: src/Components/ResetFilters/ResetFilters/ResetFilters.jsx

src/Components/CompareList/CompareList.jsx
mjoyce91: Done

src/Components/ResetFilters/ResetFilters/ResetFilters.jsx
mjoyce91: @burgwyn - I wanted to make use of a connected component which would handle the route change, and a child component to handle the presentation. Maybe they could be combined into the same file but still exist as separate classes? I named the parent/connected component `ResetFiltersConnect.jsx` but maybe there's a better naming convention.

Issue #120, opened 2017-06-19
Share Position with an External Recipient


burgwyn:
As a bidding employee,
I would like to share a position
So I can inform an external recipient of a potential position

**Acceptance Criteria**

* A visible indicator to initiate the sharing of the position
* Designate at least 1 recipient by email address
* User can share with multiple recipients
* Prompt to remind the sender the recipient is outside the department
* An email communication with plain language capturing:
  * Position Title
  * Position Number
  * Post
  * Disclaimer about sharing, etc

PR #223, opened 2017-07-20
Share position with external recipient

mjoyce91:
Adds functionality from #120 :
- Integrates with new email API endpoint to send a position by email (note: real emails are not actually sent yet, pending email server)
- Runs a regular expression to check that email ends with "state.gov" and alerts the user if their input does not end as such
- Confirmation that email was sent is alerted to user


src/Components/Share/Share.test.jsx
```diff
@@ -0,0 +1,24 @@
...
+import thunk from 'redux-thunk';
+import Share from './Share';
+
+const middlewares = [thunk];
+const mockStore = configureStore(middlewares);
+
+describe('Main', () => {
+  const api = 'http://localhost:8000/api/v1';
+
+  beforeEach(() => {
```
burgwyn: Do we need this?

src/Components/Share/ShareButton/ShareButton.jsx
```diff
@@ -0,0 +1,83 @@
...
+      recipient: '',
+      warning: false,
+      timeout: false,
+    };
+  }
+
+  changeEmail(e) {
+    const recipient = e.target.value;
+    this.setState({ recipient });
+    const warning = !(/@state.gov\s*$/.test(recipient));
```
burgwyn: We should externalize this

burgwyn: And confirm the pattern

src/reducers/share.js
```diff
@@ -0,0 +1,24 @@
+export function shareHasErrored(state = false, action) {
+  switch (action.type) {
+    case 'SHARE_HAS_ERRORED':
```
burgwyn: I like the consistency of this pattern.  Do you think it is worth tracking some additional events, like _IS_STARTING?

src/Components/Share/Share.test.jsx
mjoyce91: Removed

src/Components/Share/ShareButton/ShareButton.jsx
mjoyce91: Good point, there may be sub domains or alternate domains for valid State employee email addresses.

PR #231, opened 2017-07-21
Add Loading component

mjoyce91:


src/Components/ResultsPage/ResultsPage.jsx
```diff
@@ -21,13 +22,7 @@ class Results extends Component {
...
-    const e = this.props.hasErrored ? (
-      <span>There was an error loading the results</span>
-    ) : null;
-    const l = this.props.isLoading && !this.props.hasErrored ? (<span>Loading...</span>) : null;
-    const n = !this.props.isLoading && !this.props.hasErrored && !results.length ?
-      <span>No results with that search criteria</span> : null;
+    const { results, isLoading, hasErrored } = this.props;
     return (
       <div className="usa-grid-full">
         <div className="usa-grid-full">
```
burgwyn: Do we need both grids?  Is there an equivilant to 'row'?

mjoyce91: @burgwyn there's a `usa-width-one-whole` class, but it's not doing exactly what I'd expect. I'll spend a little more time looking into it.

burgwyn: @mjoyce91 - the class isn't a show-stopper at the moment

Issue #25, opened 2017-06-07
Tour of duty length filter


wongperson:
As a bidding employee, 
I want to be able to filter out search results of available posts based on the tour of duty
to provide posts based on the length of the potential assignment

**Acceptance criteria:**
	- The selection of the tour length filter is provided to the user to make their desired selection
	- The default selection is null, so all tour length options are displayed
	- The bidder can pick from a tour of duty length description code
	- When a tour of duty is selected, the tour of duty choice appears to remind the user of what tour length(s) have been selected
	- More than one tour of duty may be selected

**Definition of done:**
Assumptions documented
Coded work is complete
Full unit test coverage
Product owner approval

apbdevine:
Edited story to remove Tour of Duty length based on information received from DOS team. "the standard TOD is associated with a post. We do not maintain a TOD for a position. When assignments are initiated the TOD defaults to the Post's. However, the assignment TOD can be changed by selecting from the common list of TODs or creating a customized TOD."

burgwyn:
Closing

Issue #23, opened 2017-06-07
Region filter


wongperson:
As a bidding employee, 
I want to be able to manually search positions using a region filter
to provide available positions that are in a certain geographic region

**Acceptance criteria:**
	- The presentation of the region filter is provided to the user to make their desired selection
	- The default selection is null, so all regions are displayed
	- The available Foreign Service regions should be presented by regional Bureau 
	- When a region is selected, the region choice appears to remind the user of what region(s) they have selected
	- More than one region can be selected as a filter at a time

**Definition of done:**
Assumptions documented
Coded work is complete
Full unit test coverage
Product owner approval

burgwyn:
Closing

PR #234, opened 2017-07-24
Sprint 3

mjoyce91:
This PR includes functionality scheduled for Sprint 3, including:

#23 - Region Filter included on Execute Search page
#25 - Tour of Duty Filter included on Execute Search page
#29 - User has ability to clear filters from the Search Results page by returning to the Execute Search page with default filters (all unselected)
#117 - User has ability to compare positions by marking them on the Search Results and navigating to the Compare page, by a link accessible from the Results page
#120 - User can share a position with an external recipient. This is currently making a POST request to a mock endpoint on the API. No email is actually sent. However, proper HTTP responses are returned as if the email was actually sent.

Additionally, the application has been integrated with React-Redux in order to retrieve data for each page. This includes some additional abstraction using connected components. Additionally, some functionality has been externalized to individual components, such as the Language List and Loading alerts, as they are used throughout the site.


src/Constants/PropTypes.js
```diff
@@ -0,0 +1,75 @@
...
+  organization: PropTypes.string,
+  position_number: PropTypes.string,
+  is_overseas: PropTypes.boolean,
+  create_date: PropTypes.string,
+  update_date: PropTypes.string,
+  post: POST_MISSION_DATA,
+  languages: LANGUAGES,
+});
+
+export const RESULTS = PropTypes.arrayOf(
```
jseppi: `POSITION_SEARCH_RESULTS` or something might be a bit of a nicer name since `RESULTS` could be nearly anything.

src/Containers/Position/Position.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+const mapStateToProps = (state, ownProps) => ({
+  positionDetails: state.positionDetails,
+  hasErrored: state.positionDetailsHasErrored,
+  isLoading: state.positionDetailsIsLoading,
+  id: ownProps,
+});
+
+const mapDispatchToProps = dispatch => ({
+  fetchData: url => dispatch(positionDetailsFetchData(url)),
```
jseppi: A perhaps nicer pattern to use here and on other actions that use an API endpoint would be to just take the necessary parameters as arguments (like `position_number` in this case), and encapsulate the rest of the API endpoint URL in the action implementation itself. I think that would also allow you to remove the `api` prop from a lot of components, and instead keep it in a single place (that can be conditionally changed according to dev/test/staging/prod/etc environment).

src/Containers/Results/Results.test.jsx
```diff
@@ -1,60 +1,29 @@
...
-      setTimeout(() => {
-        expect(wrapper.instance().state.results.length).toBe(2);
-        done();
-      }, 0);
-    };
-    f();
+  it('is defined', () => {
+    const results = TestUtils.renderIntoDocument(<Provider store={mockStore({})}><MemoryRouter>
+      <Results api={api} />
+    </MemoryRouter></Provider>);
```
jseppi: This pattern is used quite often (in all I think) of the tests, so it might be nice to extract the shared code. 

Same with the `mockStore` setup used in this file (line 11) and elsewhere.

src/actions/results.test.js
```diff
@@ -0,0 +1,61 @@
...
+    );
+  });
+
+  it('can fetch a position', (done) => {
+    const store = mockStore({ results: [] });
+
+    const f = () => {
+      setTimeout(() => {
+        store.dispatch(actions.resultsFetchData('http://localhost:8000/api/v1/results/'));
+    // .then(do something)
```
jseppi: Not needed?

src/Containers/Home/Home.jsx
```diff
@@ -1,322 +1,72 @@
...
-    }),
-  ),
-),
+  onNavigateTo: PropTypes.func,
+  fetchData: PropTypes.func,
+  isLoading: PropTypes.bool,
+  items: ITEMS,
 };
 
 Home.defaultProps = {
```
jseppi: Seeing some prop validation errors on load:

screen shot 2017-07-25 at 1 27 00 pm

src/actions/filters.js
```diff
@@ -0,0 +1,46 @@
...
+    const responses = [];
+
+    function dispatchSuccess() {
+      if (responses.length === items.length) {
+        dispatch(filtersIsLoading(false));
+        dispatch(filtersFetchDataSuccess(responses));
+      }
+    }
+
+    items.forEach((item) => {
```
jseppi: I'm a little confused deciphering what is going on here. It looks like a new AJAX request is fired for each element of `items`, so `dispatchSuccess()` is called multiple times. Is that desired behavior? Would it maybe make more sense to wait for all of the `.get` requests to finish before the final success/error dispatch?

Something like:

```js
const itemPromises = items.map((item) => axios.get('...').then((response) =>  {data: response.data, item: itemFilter.item }));
Promise.all(itemPromises)
  .then((allResults) => { 
    responses = allResults;
    dispatchSuccess();
  })
  .catch(() => dispatch(filtersHasErrored(true)));
```

src/Components/FavoritesButton/FavoritesButton.jsx
```diff
@@ -1,12 +1,15 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
+import { localStorageFetchValue, localStorageToggleValue } from '../../utilities';
+import { EMPTY_FUNCTION } from '../../Constants/PropTypes';
 
 class FavoritesButton extends Component {
   constructor(props) {
     super(props);
     this.state = {
```
jseppi: Should this be instead `connect()`'d to the Redux store?

src/Components/FavoritesButton/FavoritesButton.jsx
```diff
@@ -16,57 +19,54 @@ class FavoritesButton extends Component {
...
   }
 
   getSavedState() {
     return this.state.saved;
   }
 
-  initSetSaved(val) {
-    const saved = val;
-    this.setState({ saved });
+  exceedsLimit() {
```
jseppi: can be simplified to just `return this.state.len >= this.props.limit;`

src/Components/Filters/Filters.test.jsx
```diff
@@ -0,0 +1,256 @@
...
+      }, 0);
+    };
+    f();
+  });
+
+  it('should disable search if less than two filters are selected', (done) => {
+    wrapper = shallow(<Filters api={api} items={items} />, { context });
+    const f = () => {
+      setTimeout(() => {
+    // no filters are initially set, so should return true
```
jseppi: style: indentation of comments and code seems off from here through the end of this file

src/Components/Filters/Filters.jsx
```diff
@@ -0,0 +1,308 @@
...
+    };
+  }
+
+  changeProficiency(prof, val) {
+    const { proficiency } = this.state;
+    proficiency[prof] = val;
+    this.setState({ proficiency });
+  }
+
+  createQueryString() {
```
jseppi: Consider extracting the logic here for easier testing. I didn't notice any tests for this method, and it has a good amount of logic so it should be thoroughly tested.

src/Components/Filters/Filters.test.jsx
```diff
@@ -0,0 +1,256 @@
...
+        done();
+      }, 0);
+    };
+    f();
+  });
+
+  it('should be able to submit a search', (done) => {
+    wrapper = shallow(<Filters api={api} items={items} />, { context });
+    const f = () => {
+      setTimeout(() => {
```
jseppi: What are the uses of `setTimeout(..., 0)` for in this test case and others in this file?

src/Components/LanguageList/LanguageList.test.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import LanguageList from './LanguageList';
+
+describe('DetailsComponent', () => {
```
jseppi: Should probably have a test that it is actually rendering a `span` containing the `choice` text for each `language`.

https://www.npmjs.com/package/enzyme-to-json might be a nice way to test rendering without manually typing out all the expected DOM for each component under test.

src/actions/filters.js
mjoyce91: So the `dispatchSuccess()` function is really just checking that all the AJAX requests for the filters have been returned by comparing `responses.length` to `items.length` - it doesn't actually dispatch a success until those lengths match, at which point `dispatch(filtersFetchDataSuccess(responses))` is called, which is what actually fires the action. Your's looks cleaner and does the same thing without the extra logic.

src/utilities.js
```diff
@@ -8,4 +8,42 @@ export const ajax = url => (
...
+        JSON.stringify(existingArray));
+  } else {
+    existingArray.push(value);
+    localStorage.setItem(key,
+        JSON.stringify(existingArray));
+  }
+}
+
+export function validStateEmail(email) {
+  const valid = (/@state.gov\s*$/.test(email));
```
mjoyce91: @jseppi - I'll use your regex for now, and if we see find a need in other areas for validation, we can bring in this library. The best solution may become more clear once we actually have a mail server.

package.json
```diff
@@ -62,7 +70,7 @@
   "scripts": {
     "start": "node scripts/start.js",
     "build": "node scripts/build.js",
-    "test": "node scripts/test.js --env=jsdom --no-cache --coverage"
+    "test": "node scripts/test.js --env=jsdom --no-cache --coverage --runInBand"
```
mjoyce91: @jseppi - tests were running fine locally, but we were hitting memory limits on CircleCI. I found that adding this parameter resolved the memory issues. Take a look at these two builds to see [before](https://circleci.com/gh/18F/State-TalentMAP/197) and [after](https://circleci.com/gh/18F/State-TalentMAP/198)

src/Components/Share/ShareButton/ShareButton.jsx
```diff
@@ -0,0 +1,85 @@
...
+    const sendingText = isSending ? 'Sending...' : null;
+    const err = hasErrored ? 'Sorry, there was a problem sharing this.' : null;
+    const sent = (response && !hasErrored && !isSending && timeout) ? 'Sent!' : null;
+    return (
+      <form onSubmit={e => this.share(e)}>
+        <label htmlFor="share-input">Share this position by email</label>
+        <input
+          id="share-input"
+          name="input-type-text"
+          type="text"
```
burgwyn: Nothing makes me more upset than using a mobile device, encountering an obviously numeric field and being forced into using the text keyboard.

src/Components/Filters/Filters.test.jsx
mjoyce91: @jseppi - I found that these were needed when the test relies on data returned by `axios-mock-adapter`.

src/Components/FavoritesButton/FavoritesButton.jsx
mjoyce91: @jseppi - first I'll need to integrate local storage with Redux.

src/Components/Filters/Filters.jsx
mjoyce91: @jseppi - it's implicitly tested in the `it can create a query string` test, but could use some more thorough testing.

PR #273, opened 2017-07-27
LanguageList - Adding Snapshot Tests

burgwyn:
Moving unit tests to use enzyme and snapshots


jseppi:
How are you liking the snapshot tests? I've found them useful on other projects.

PR #274, opened 2017-07-28
Login

mjoyce91:
This PR introduces login using redux-saga and the `/accounts/token/` endpoint. All routes must pass a `checkAuthorization` function which checks for the existence of a token in the browser's local storage. If the user is logged in, they can navigate the site normally. If not, they are redirected to the `/login` route.


mjoyce91:
Adding @jseppi as a reviewer as well since this is a bigger feature.

src/api.js
```diff
@@ -0,0 +1,2 @@
+const api = 'http://localhost:8000/api/v1';
```
burgwyn: This is fine for now.  Let's think about longer term.

src/index-sagas.js
```diff
@@ -0,0 +1,7 @@
+import LoginSaga from './login/sagas';
+
+export default function* IndexSaga() {
```
burgwyn: GENERATORS!!! https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator

src/lib/check-auth.js
```diff
@@ -0,0 +1,27 @@
...
+function checkAuthorization(dispatch) {
+  // attempt to grab the token from localstorage
+  const storedToken = localStorage.getItem('token');
+
+  // if it exists
+  if (storedToken) {
+    // parse it down into an object
+    const token = JSON.parse(storedToken);
+
+    // TODO include check for token expiration
```
burgwyn: Make an issue if we don't have one

src/notifications/Errors.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// Iterate over each error object and print them
+// in an unordered list
+const Errors = (props) => {
+  const { errors } = props;
+  return (
+    <div>
```
burgwyn: Should we consider something like USWDS Alerts? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Generator

src/notifications/Messages.jsx
```diff
@@ -0,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+// Iterate over each message object and print them
+// in an unordered list
+const Messages = (props) => {
+  const { messages } = props;
+  return (
+    <div>
```
burgwyn: USWDS Alert?

src/notifications/Errors.jsx
mjoyce91: @burgwyn Did you mean to reference this? https://standards.usa.gov/components/alerts/

src/login/sagas.js
```diff
@@ -0,0 +1,106 @@
...
+  }
+
+  // return the token for health and wealth
+  return token;
+}
+
+// Our watcher (saga).  It will watch for many things.
+function* loginWatcher() {
+  // Check if user entered already logged in or not
+  while (true) {
```
jseppi: I'm not actually familiar with `redux-sagas`, but I've heard good things. Is this `while(true)` a common pattern?

Also this line is getting an eslint warning (`no-constant-condition`) that you can `eslint-disable-line` if this pattern is good.

src/login/sagas.js
```diff
@@ -0,0 +1,106 @@
...
+
+const loginUrl = `${api}/accounts/token/`;
+
+export const errorMessage = { message: null };
+
+export function changeErrorMessage(e) {
+  errorMessage.message = e;
+}
+
+export function loginApi(email, password) {
```
jseppi: Heads-up: When I spun up this branch locally, I was **not** able to login with my user's email address, but I was able to login with my username.

src/notifications/Errors.jsx
mjoyce91: Made an Alerts component in the `alert-messages` branch. Will wait for James' review before merging that into this branch/PR.

src/login/sagas.js
mjoyce91: Good catch. This is expected as far as I'm concerned, and is really just a wording decision I made. The API is expecting a username, not an email address. I can update the fields to reference "username" instead of "email".

jseppi: Ok, then I think the UI needs to be modified as well: https://github.com/18F/State-TalentMAP/pull/274/files#diff-829fd2601590ad4f7baf99d8d3ca3d11R46

src/login/sagas.js
mjoyce91: Seems to be a common approach: https://github.com/redux-saga/redux-saga/blob/1230d2299367d50a94135a0d0a036a7f5b6775ae/docs/recipes/README.md

src/login/sagas.js
mjoyce91: Done

src/Containers/Main/Main.jsx
```diff
@@ -25,27 +36,58 @@ function configureStore(initialState) {
...
+          <Switch {...props}>
+            <Route
+              path="/login"
+              component={() => (!isAuthorized() ?
+              (<Login {...props} />) : (<Redirect to="/" />))}
+            />
+            <Route
+              exact
+              path="/"
+              component={() => (isAuthorized() ?
```
jseppi: Right after logging in, I am seeing a dev console error about setting state during a state transition:

screen shot 2017-07-28 at 1 24 34 pm

I'm not yet familiar enough with sagas to know what is causing this because I'm not seeing an explicit `setState` call anywhere in this PR, but it seems like something isn't totally right.

mjoyce91: Must have something to do with the auth check occurring on the routes. Here's a current route that causes this warning:

```
<Route
              exact
              path="/"
              component={() => (isAuthorized() ?
                (<Home {...props} />) : (<Redirect to="/login" />))}
            />
```
While this does not:
```
<Route
              exact
              path="/"
              component={() => <Home {...props} />}
            />
```
I'll keep looking into it.

mjoyce91: @jseppi the issue seemed to be related to the conditional rendering of either the component or a `<Redirect ...>` based on `isAuthorized()`. I moved the `isAuthorized` check to the container level and the warning has gone away.

src/login/sagas.js
jseppi: Cool, could you add `// eslint-disable-line no-constant-condition` to this line?

mjoyce91: Done

src/login/sagas.test.js
```diff
@@ -0,0 +1,72 @@
...
+import MockAdapter from 'axios-mock-adapter';
+import loginWatcher, { changeErrorMessage, errorMessage, loginApi } from './sagas';
+
+describe('login functions', () => {
+  it('can log in and set the client', () => {
+    const mockAdapter = new MockAdapter(axios);
+    mockAdapter.onPost('http://localhost:8000/api/v1/accounts/token/').reply(200,
+      { token: '12345' },
+    );
+    return expectSaga(loginWatcher)
```
jseppi: Look like every instance of `expectSaga` produces a warning when running jest: `[WARNING]: Saga exceeded async timeout of 250ms`.  Any idea what that is about?

I even tried increasing the timeout (`expectSaga.DEFAULT_TIMEOUT = 2000;`) and the warning is still produced.

jseppi: Ah-ha! I guess it is expected because the `loginWatcher` runs in an infinite loop, so hitting the timeout is necessary to stop the test. It's kinda documented in the bottom of the this section: https://github.com/jfairbank/redux-saga-test-plan/blob/a4c3c8950c206abc2e71de2673f8b64dd9198223/docs/integration-testing/timeout.md#adjusting-timeout

src/login/index.jsx
```diff
@@ -0,0 +1,118 @@
...
+      handleSubmit, // remember, Redux Form injects this into our props
+      login: {
+        requesting,
+        successful,
+        messages,
+        errors,
+      },
+    } = this.props;
+
+    // Remember, Redux Form passes the form values to our handler
```
jseppi: I appreciate all the explanatory comments!

src/login/index.jsx
```diff
@@ -0,0 +1,118 @@
...
+          <div className="auth-messages">
+            {/* As in the signup, we're just using the message and error helpers */}
+            {!requesting && !!errors.length && (
+            <Errors message="Failure to login due to:" errors={errors} />
+          )}
+            {!requesting && !!messages.length && (
+            <Messages messages={messages} />
+          )}
+            {requesting && <div>Logging in...</div>}
+            {!requesting && !successful}
```
jseppi: Does this line do anything?

src/login/index.jsx
```diff
@@ -0,0 +1,118 @@
...
+                component="input"
+              />
+              <input type="submit" value="Sign in" />
+            </fieldset>
+          </form>
+          <div className="auth-messages">
+            {/* As in the signup, we're just using the message and error helpers */}
+            {!requesting && !!errors.length && (
+            <Errors message="Failure to login due to:" errors={errors} />
+          )}
```
jseppi: formatting is a little hard to read here. maybe something like:

```jsx
{
  !requesting && !!errors.length &&
  (<Errors message="Failure to login due to:" errors={errors} />)
}
```
?

src/login/index.jsx
mjoyce91: Conveniently, the example I used already had them. Though I did have to change a number of them based on our implementation.

src/login/index.jsx
mjoyce91: I'm working on a branch that uses USWDS alerts for this, but I'll double check line spacing for readability.

src/login/index.jsx
mjoyce91: Probably an artifact of catching invalid login before the use of `errors`. Removing.

PR #277, opened 2017-07-28
Externalize and style the footer

mjoyce91:
Moved footer to its own component and added non-pixel-perfect style from the wireframe. Makes use of snapshot testing.


src/Components/Footer/Footer.test.jsx
```diff
@@ -0,0 +1,18 @@
...
+
+describe('DetailsComponent', () => {
+  let footer = null;
+
+  it('is defined', () => {
+    footer = shallow(<Footer />);
+    expect(footer).toBeDefined();
+  });
+
+  it('takes props', () => {
```
jseppi: Name of test should be something like "matches snapshot"

src/Components/Footer/Footer.test.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJSON from 'enzyme-to-json';
+import Footer from './Footer';
+
+describe('DetailsComponent', () => {
```
jseppi: Should be FooterComponent

src/Components/Footer/Footer.test.jsx
```diff
@@ -0,0 +1,18 @@
...
+  let footer = null;
```
jseppi: No reason to instantiate the variable here. Can just do it in 'it' cases

mjoyce91:
@burgwyn @jseppi - thanks for feedback. See updates.

Issue #247, opened 2017-07-26
Resolve invalid PropTypes with filters


mjoyce91:
```
console.error node_modules/fbjs/lib/warning.js:36
      Warning: Failed prop type: Invalid prop `items[0].data[0]` of type `object` supplied to `Filters`, expected an array.
          in Filters (at Filters.test.jsx:118)
```

mjoyce91:
@burgwyn - I'm gonna take a stab at this one but may re-assign to you.

burgwyn:
Closing as this was completed

PR #278, opened 2017-07-31
Removing redundant `PropTypes.arrayOf`

burgwyn:


mjoyce91:
Nice, this resolves #247 👍

Issue #219, opened 2017-07-19
COLA filter


burgwyn:
As a bidding employee,
I want to be able to add search criteria to include positions with COLA
for opportunities with COLA in my results list

Acceptance criteria:
- The bidder is able to indicate if they would like to include COLA opportunities
- By default, COLA opportunities are included in the positions results list
- The bidder must be able to read text content that asks if the bidder would like to include COLA opportunities (or some form of instruction)
- When indicated, the search includes a COLA filter

Definition of done:
Assumptions documented
Coded work is complete
Full unit test coverage
Product owner approval

mjoyce91:
@burgwyn  : @afrasier and I are looking at the new filters for this sprint. They all follow the logic of "Choose true to show all positions, choose false to show positions with no XYZ differential". Filtering on True is easy since you don't need to add any new parameters. but filtering on False is trickier since there are differentials with value `0` but also positions with `null` posts. I'm assuming we'd want to show both the positions with `0` as well as the positions with a `null` post. Should @afrasier build something out to account for that, or do we not want to worry about it since positions in the real data will always be associated with a post?

burgwyn:
@mjoyce91 - I think we will be ok.  I take the story as "show me positions with a value > 0" for the true path.  The false path would be "don't include X filter in the search".

mjoyce91:
@burgwyn - that clears things up. Thanks!

esgoodman:
@burgwyn @mjoyce91 @krisdelacruz I think we should deprioritize these filters given results from usability tests with bidders.

burgwyn:
Closing

Issue #215, opened 2017-07-19
Domestic filter


burgwyn:
As a bidding employee,
I want to be able to manually search positions using a domestic filter
to provide available positions that are in the US

Acceptance criteria:
- The presentation of the domestic filter is provided to the user to make their desired selection
- The default selection is null, so all positions are displayed
- When a region bureau is selected, the region bureau choice appears to remind the user of what region bureau(s) they have selected
- More than one region bureau can be selected as a filter at a time

Definition of done:
Assumptions documented
Coded work is complete
Full unit test coverage
Product owner approval

mjoyce91:
@burgwyn - looks like the description needs to be updated on this one. Domestic vs region bureau / copy and paste.

burgwyn:
Closing

Issue #31, opened 2017-06-07
Post differential filter


burgwyn:
As a bidding employee,
I want to be able to add search criteria to include positions with post differential
for opportunities with post differential pay in my results list

Acceptance criteria:
- The bidder is able to indicate if they would like to include post differential pay opportunities
- By default, post differential pay opportunities are included in the positions results list
- The bidder must be able to read text content that asks if the bidder would like to include post differential pay (or some form of instruction)
- When indicated, the search includes a post differential pay filter

Definition of done:
Assumptions documented
Coded work is complete
Full unit test coverage
Product owner approval

afrasier:
@mjoyce91 `/api/v1/positions/post__differential_rate__gt=0` will return positions with only posts having differential rate > 0

esgoodman:
@burgwyn @mjoyce91 @krisdelacruz I think we should deprioritize these filters given results from usability tests with bidders.

burgwyn:
Closing

Issue #26, opened 2017-06-07
Danger pay filter option


wongperson:
As a bidding employee, 
I want to be able to add search criteria to include positions with danger pay
for opportunities with danger pay in my results list

**Acceptance criteria:**
	- The bidder is able to indicate if they would like to include danger pay opportunities
	- By default, danger pay opportunities are included in the positions results list
	- The bidder must be able to read text content that asks if the bidder would like to include danger pay opportunities (or some form of instruction)
	- When indicated, the search includes a danger pay filter

**Definition of done:**
Assumptions documented
Coded work is complete
Full unit test coverage
Product owner approval

afrasier:
@mjoyce91 `/api/v1/positions/post__danger_pay__gt=0` will return positions with only posts having danger pay > 0

burgwyn:
Closing

PR #279, opened 2017-07-31
Sprint 4 filters - COLA, Post Differential, Danger Pay, Domestic

mjoyce91:
This PR introduces four new filters - COLA, Post Differential, Danger Pay, Domestic. These filters are all boolean options and do not require any AJAX requests. Therefore, they are structured slightly differently than previous filters by making use of a `bool: true` property which determines the presentation and data-fetching of the filter. This covers the following epics:

#219 - COLA filter
#31 - Post Differential filter
#26 - Danger Pay filter
#215 - Domestic filter


mjoyce91:
@burgwyn good idea - maybe once the search design is more finalized?

PR #296, opened 2017-08-02
Alert messages

mjoyce91:
This PR introduces an Alert component which can be used for various types of alert messages throughout the site. It makes use of [USWDS Alerts](https://standards.usa.gov/components/alerts/). A title and body can be passed as props, as well as the type (success, warning, error, etc.) to determine the style. The only place it is implemented in this PR is within the login screen to show status of login ("logging in" or "login error").


src/Components/Alert/Alert.jsx
```diff
@@ -0,0 +1,30 @@
...
+  type: PropTypes.string,
+  title: PropTypes.string.isRequired,
+  messages: PropTypes.arrayOf(
+    PropTypes.shape({
+      body: PropTypes.string,
+    })),
+};
+
+Alert.defaultProps = {
+  type: 'success', // should be one of the USWDS alert types - https://standards.usa.gov/components/alerts/
```
burgwyn: Maybe we default to info?

src/Components/Alert/Alert.jsx
```diff
@@ -0,0 +1,30 @@
...
+      <h3 className="usa-alert-heading">{title}</h3>
+      <p className="usa-alert-text">
+        {messages.map(message => message.body)}
+      </p>
+    </div>
+  </div>
+);
+
+Alert.propTypes = {
+  type: PropTypes.string,
```
burgwyn: Can we use an enum?

src/Components/Alert/Alert.jsx
```diff
@@ -0,0 +1,30 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const Alert = ({ type, title, messages }) => (
+  // 'type' is injected into the class name
+  <div className={`usa-alert usa-alert-${type}`}>
```
burgwyn: The error type requires an ARIA `role="alert"` https://standards.usa.gov/components/alerts/

src/Components/Alert/Alert.jsx
```diff
@@ -0,0 +1,31 @@
...
+import PropTypes from 'prop-types';
+
+const Alert = ({ type, title, messages }) => (
+  // 'type' is injected into the class name
+  // type 'error' requires an ARIA role
+  <div className={`usa-alert usa-alert-${type}`} role={(type === 'error') ? 'alert' : null}>
+    <div className="usa-alert-body">
+      <h3 className="usa-alert-heading">{title}</h3>
+      <p className="usa-alert-text">
+        {messages.map(message => message.body)}
```
jseppi: There will be no line breaks between messages as this is written. Perhaps move the `<p>` tag inside the `.map`:

```jsx
{messages.map(message => <p className="usa-alert-text">{message.body}</p>)}
```

I guess technically a `<ul>` with each message inside of an `<li>` would be more correct, but that will look weird in the case of `message.length === 1`. Though you could of course add some simple logic to take care of that as well.

burgwyn: I've made the move to push this into separate `<p>` tags.  I see your point on `<li>` but we may want to wait to see some design guidance before going there.

src/Components/Alert/Alert.jsx
```diff
@@ -0,0 +1,33 @@
...
+import PropTypes from 'prop-types';
+
+const Alert = ({ type, title, messages }) => (
+  // 'type' is injected into the class name
+  // type 'error' requires an ARIA role
+  <div className={`usa-alert usa-alert-${type}`} role={(type === 'error') ? 'alert' : null}>
+    <div className="usa-alert-body">
+      <h3 className="usa-alert-heading">{title}</h3>
+      {messages.map((message, index) =>
+        (<p className="usa-alert-text" key={index}>{/* eslint-disable react/no-array-index-key */}
```
jseppi: Hrm, using the loop index could potentially lead to things not being rendered, I think. This article seems to explain it well: https://medium.com/@robinpokorny/index-as-a-key-is-an-anti-pattern-e0349aece318

Perhaps use `shortid` as it recommends?

burgwyn: I knew `index` was sub-optimal, but the article really helped dig into why.  I've added `shortid`.

PR #297, opened 2017-08-02
Header

mjoyce91:
This PR adds the styled header, which includes an Account Dropdown using [react-simple-dropdown](https://github.com/Fauntleroy/react-simple-dropdown).

This PR covers #238 - Add style from mockups to front-end header


src/Components/Header/Header.jsx
```diff
@@ -93,74 +86,6 @@ export class Header extends Component {
             <button className="usa-nav-close">
               <img src="https://unpkg.com/uswds@1.0.0/dist/img/close.svg" alt="close" />
```
burgwyn: We've got `font-awesome` and an updated version of `uswds` in another pull request.  We should consider using one of those and avoid the external request.

src/Components/Header/Header.jsx
```diff
@@ -172,11 +97,17 @@ export class Header extends Component {
...
+                </li>
+                <li>
+                  <Link to="/">Home</Link>
+                </li>
+                <li>
+                  <a href="https://github.com/18F/State-TalentMAP">About</a>
+                </li>
                 <li>
-                  <a href="/#">Secondary link</a>
+                  <a href="https://github.com/18F/State-TalentMAP/issues">Feedback</a>
```
burgwyn: I know these are placeholders, but good call 👍

src/Components/Header/Header.test.jsx
```diff
@@ -1,5 +1,6 @@
 import React from 'react';
 import { shallow } from 'enzyme';
+import toJSON from 'enzyme-to-json';
 import { Header } from './Header';
 
 describe('Main', () => {
```
burgwyn: 🍝 - `Header`

src/Components/AccountDropdown/AccountDropdown.jsx
```diff
@@ -0,0 +1,56 @@
...
+        <DropdownTrigger href="/#">
+          <img
+            alt="avatar"
+            className="account-dropdown--avatar"
+            src="/assets/img/avatar.png"
+          />
+          <span className="account-dropdown--name">User</span>
+        </DropdownTrigger>
+        <DropdownContent>
+          <div className="account-dropdown--identity account-dropdown--segment">
```
burgwyn: Are these class names `account-dropdown--identity` part of the library?  I can't explain it, but the `--` really gets to me

mjoyce91: Yeah, unfortunately these are part of the library.

src/Components/Header/Header.jsx
mjoyce91: Good call, think this was part of the header code I copied directly from their site. There are some static assets within the project I can reference instead of this external link.

src/Components/AccountDropdown/AccountDropdown.jsx
```diff
@@ -0,0 +1,56 @@
...
+  }
+
+  render() {
+    return (
+      <Dropdown className="account-dropdown">
+        <DropdownTrigger href="/#">
+          <img
+            alt="avatar"
+            className="account-dropdown--avatar"
+            src="/assets/img/avatar.png"
```
mjoyce91: Right, we'll want to customize a lot more with profile data like name, picture, etc.

src/Components/Header/Header.jsx
burgwyn: Yeah, I saw it in the USWDS site, though it looks like [they've pulled back from using that CDN](https://github.com/18F/web-design-standards/issues/1931).  I was debating creating an issue for [Content Security Policy](https://developer.mozilla.org/en-US/docs/Web/HTTP/CSP) and now have #302

src/Components/AccountDropdown/AccountDropdown.test.jsx
```diff
@@ -0,0 +1,21 @@
...
+
+describe('AccountDropdown', () => {
+  it('is defined', () => {
+    const accountDropdown = shallow(<AccountDropdown />);
+    expect(accountDropdown).toBeDefined();
+  });
+
+  it('can click the logout link', () => {
+    const accountDropdown = shallow(<AccountDropdown />);
+    accountDropdown.find('[to="login"]').simulate('click');
```
mjoyce91: Added sinon to spy the `logout()` function and test that it was called.

Issue #119, opened 2017-06-19
Share a Position with Another Employee


burgwyn:
As an employee,
I would like to share a position
So I can inform another user of a potential position

**Acceptance Criteria**

* A visible indicator reflecting the sharing of the position
* User must designate 1 recipient
* Confirmation that the position was shared
* A notification with plain language capturing:
  * Name of the sender is 'the system'
  * Position Title
  * Position Number
  * Post
  * A short description of the position
  * A hyperlink to the position details
* The share would be recorded and stored 
  * Duration is indefinite but subject to record retention requirements
* The system would send an email message to indicate that there is a message, the message body would live inside the system

burgwyn:
Closing

PR #304, opened 2017-08-03
Internal sharing

mjoyce91:
This PR addresses #119 - Share a position with another employee. A radio button is now displayed above the email input to allow the user to choose between an Internal and an External share. Additionally, if there is an error message from the server, such as "User does not exist", this is displayed to the user instead of a generic error message.


src/Components/Share/ShareButton/ShareButton.jsx
```diff
@@ -29,40 +30,70 @@ class ShareButton extends Component {
...
+            />
+            <label htmlFor="internal">Internally</label>
+          </fieldset>
+          <input
+            id="share-input"
+            name="input-type-text"
+            type="text"
+            value={this.state.recipient}
+            onChange={e => this.changeEmail(e)}
+            placeholder="Receipient's email address"
```
burgwyn: Typo

src/Components/Share/ShareButton/ShareButton.jsx
```diff
@@ -29,40 +30,70 @@ class ShareButton extends Component {
...
+          <button className={(recipient.length && !isSending) ? null : 'usa-button-disabled'} disabled={recipient.length ? null : true} id="share-button">
           Share
         </button>
-        {warning && recipient.length ? 'This is not a state.gov email. Send with caution.' : null}
-        <br />
-        {sendingText} {err} {sent}
-      </form>
+          {warning && recipient.length ? 'This is not a state.gov email. Send with caution.' : null}
+          <br />
+          {sendingText} {err} {sent}
```
burgwyn: Could all three or more than one have values simultaneously?

src/Components/Share/ShareButton/ShareButton.jsx
```diff
@@ -29,40 +30,70 @@ class ShareButton extends Component {
...
+          </fieldset>
+          <input
+            id="share-input"
+            name="input-type-text"
+            type="text"
+            value={this.state.recipient}
+            onChange={e => this.changeEmail(e)}
+            placeholder="Receipient's email address"
+          />
+          <button className={(recipient.length && !isSending) ? null : 'usa-button-disabled'} disabled={recipient.length ? null : true} id="share-button">
```
burgwyn: I'm not sure about using `recipient.length` in validations.  Could it throw false positives?  Could we flag some of this inside `share(e)`?

src/Components/Share/ShareButton/ShareButton.jsx
mjoyce91: 🔠

src/Components/Share/ShareButton/ShareButton.jsx
mjoyce91: All I'm really checking for here is that _some_ text has been entered and a share is not currently being sent -- if so, the button is disabled. I think what @jseppi and I discussed was that a lot more validation will ultimately take place on the server, and we can adjust the front-end accordingly when that happens.

src/Components/Share/ShareButton/ShareButton.jsx
mjoyce91: You're right - that is possible. I updated the action to set the state of `hasErrored`, `isSending`, and `Success` in such a way that only one of the three is ever true.

PR #309, opened 2017-08-04
Moving routes into a new container

burgwyn:
Decomposing `Main` to make the underlying router behavior easier to test

### DIFF
--- a/src/Containers/Routes/Routes.test.jsx
+++ b/src/Containers/Routes/Routes.test.jsx
@@ -11,9 +11,15 @@ const mockStore = configureStore(middlewares);
 
 describe('Routes', () => {
   const api = 'http://localhost:8000/api/v1';
-  it('is defined', () => {
+  it('handles a home route', () => {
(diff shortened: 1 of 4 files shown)


src/Containers/Routes/Routes.test.jsx
```diff
@@ -0,0 +1,48 @@
...
+import { MemoryRouter } from 'react-router-dom';
+import { Provider } from 'react-redux';
+import configureStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+import Routes from './Routes';
+
+const middlewares = [thunk];
+const mockStore = configureStore(middlewares);
+
+describe('Routes', () => {
```
mjoyce91: We should test the `/login` route too. Looks like in order to test it, the Login component will either need to be passed a `login` prop, or given the `defaultProp` of:

```
login: {
    requesting: false,
    successful: false,
    messages: [],
    errors: [],
  },
```

mjoyce91:
🔐

PR #301, opened 2017-08-02
Position Details Redesign

mjoyce91:
This PR includes the redesign of the Position Details page described in #256 .

PR #311, opened 2017-08-04
Position details data point

mjoyce91:
Adds a more granular component for the data points inside the PositionDetailsItem component. Based on feedback from @burgwyn in https://github.com/18F/State-TalentMAP/pull/301

### DIFF
--- a/src/Components/PositionDetailsDataPoint/PositionDetailsDataPoint.jsx
+++ b/src/Components/PositionDetailsDataPoint/PositionDetailsDataPoint.jsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const PositionDetailsDataPoint = ({ title, description }) => (
+  <div className="usa-width-one-fourth position-details-description-section">
+    <div className="title">{title}</div>
+    <div className="description">{description}</div>
+  </div>
+);
+
+PositionDetailsDataPoint.propTypes = {
+  title: PropTypes.string.isRequired,
+  description: PropTypes.oneOfType([
+    PropTypes.string,
+    PropTypes.number,
+    PropTypes.element,
+  ]).isRequired,
+};
+
+export default PositionDetailsDataPoint;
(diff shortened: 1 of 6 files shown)


src/Components/PositionDetailsDataPoint/PositionDetailsDataPoint.jsx
```diff
@@ -0,0 +1,20 @@
...
+const PositionDetailsDataPoint = ({ title, description }) => (
+  <div className="usa-width-one-fourth position-details-description-section">
+    <div className="title">{title}</div>
+    <div className="description">{description}</div>
+  </div>
+);
+
+PositionDetailsDataPoint.propTypes = {
+  title: PropTypes.string.isRequired,
+  description: PropTypes.oneOfType([
```
burgwyn: I like what you're thinking, maybe we go with `PropTypes.node`? - https://facebook.github.io/react/docs/typechecking-with-proptypes.html

src/Components/PositionDetailsDataPoint/PositionDetailsDataPoint.test.jsx
```diff
@@ -0,0 +1,38 @@
...
+  });
+
+  it('handles numeric props', () => {
+    wrapper = shallow(
+      <PositionDetailsDataPoint title="Title" description={1} />,
+    );
+    expect(wrapper.instance().props.description).toBe(1);
+  });
+
+  it('handles element props', () => {
```
burgwyn: If we do `PropTypes.node` then we may want to add some other types

mjoyce91:
Changed prop type to `node` and updated tests accordingly.

src/Components/PositionDetailsDataPoint/PositionDetailsDataPoint.test.jsx
```diff
@@ -35,4 +35,12 @@ describe('PositionDetailsDataPoint', () => {
     );
     expect(wrapper.instance().props.description).toBe(description);
   });
+
+  it('handles an array of different prop types', () => {
```
burgwyn: ‼️

PR #312, opened 2017-08-04
Handle no results on the Results page using the Alerts component

mjoyce91:


src/Components/ResultsPage/ResultsPage.jsx
```diff
@@ -38,14 +39,21 @@ class Results extends Component {
         </div>
         <div className="usa-grid-full">
           {
-            results.length ?
+            !isLoading && !!results.length &&
```
burgwyn: `!!`? ‼️

mjoyce91: A "0" is rendered on the page when `results.length` is 0, but is simply `true` or `false` with `!!`.

burgwyn: TMYK🌠

PR #320, opened 2017-08-04
Fix issues found in PR #319

mjoyce91:
Code climate found issues with new lines and interactive elements in #319. This PR resolves those issues.

### DIFF
--- a/src/Components/PositionTitle/PositionTitle.jsx
+++ b/src/Components/PositionTitle/PositionTitle.jsx
@@ -1,48 +1,59 @@
-import React from 'react';
+import React, { Component } from 'react';
 import FontAwesome from 'react-fontawesome';
 import { POSITION_DETAILS } from '../../Constants/PropTypes';
 
-const PositionTitle = ({ details }) => (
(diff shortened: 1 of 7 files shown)


src/Components/PositionTitle/PositionTitle.jsx
```diff
@@ -1,48 +1,59 @@
...
-          <strong>Post website:</strong> <a href="https://www.state.gov">www.state.gov</a>
-        </div>
-        <div className="usa-width-one-half position-details-header-body">
-          <strong>Point of Contact:</strong> <a href="tel:222-222-2222">222-222-2222</a>
+class PositionTitle extends Component { // eslint-disable-line
+  render() {
+    const { details } = this.props;
+    function goBack(e) {
+      if (e.keyCode === 13 || e === 'click') {
+        window.history.back();
```
burgwyn: Can we use a method within our Router implementation to keep state?

mjoyce91: How about using the `history`'s [goBack() function](https://medium.com/@dkerrious/react-router-v4-1b04ec250c24#cd6a)?

PR #328, opened 2017-08-08
API Fixes

mjoyce91:
Update front-end based on API changes made in https://github.com/18F/State-TalentMAP-API/pull/69 , including the use of `/orgpost` instead of `/post`, and some property name changes made to the API such as the Post `location` property instead of `description`.

### DIFF
--- a/src/Components/PositionDetailsItem/__snapshots__/PositionDetailsItem.test.jsx.snap
+++ b/src/Components/PositionDetailsItem/__snapshots__/PositionDetailsItem.test.jsx.snap
@@ -29,9 +29,7 @@ exports[`PositionDetailsItem matches snapshot 1`] = `
               <Link
                 replace={false}
                 to="/post/162"
-              >
-                MASERU, LESOTHO
-              </Link>
+              />
             }
             title="POST"
           />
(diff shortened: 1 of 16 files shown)


src/Constants/SystemMessages.js
```diff
@@ -6,3 +6,4 @@ export const NO_COLA = 'None listed';
 export const NO_TOUR_OF_DUTY = 'None listed';
 export const NO_BUREAU = 'None listed';
 export const NO_ORG = 'None listed';
+export const NO_RR = 'None listed';
```
burgwyn: Can we do `NO_REST_RELAXATION`?

src/Components/PositionDetailsItem/__snapshots__/PositionDetailsItem.test.jsx.snap
```diff
@@ -29,9 +29,7 @@ exports[`PositionDetailsItem matches snapshot 1`] = `
               <Link
                 replace={false}
                 to="/post/162"
-              >
-                MASERU, LESOTHO
-              </Link>
+              />
```
burgwyn: Does our test object have the `location` set?

mjoyce91: Good catch. Just updated.

src/Constants/SystemMessages.js
mjoyce91: Updated

Issue #327, opened 2017-08-08
Externalize USWDS header accordion to its own component


mjoyce91:
The portion of the header which includes the accordion with the "An official website... Here's how you know" text should be externalized to its own component.

Issue #326, opened 2017-08-08
Add sinon spies to tests


mjoyce91:
For functions that don't have a clear output to test against, add a spy to the function to test that it was `calledOnce`.

Issue #325, opened 2017-08-08
Reference images internally instead of externally


mjoyce91:
Images should be internally referenced instead of from CDNs, such as `<img src="https://unpkg.com/uswds@1.0.0/dist/img/favicons/favicon-57.png" alt="U.S. flag" />` in the header. Doing so will make use of some of the optimizations provided by webpack, as well as ensure that all site elements are accessible within a restricted network.

Issue #324, opened 2017-08-08
Remove Login's deprecated Error and Message components


mjoyce91:
Error and alert messages are handled by the `Alert` component, so the `Error` and `Message` components are no longer needed.

PR #357, opened 2017-08-09
Sprint 4 Fixes - Round 1

mjoyce91:
Many, but not all, fixes from Sprint 4 feedback, including:

- #325 - Reference images internally
- #323 - Abbreviate long css class names
- #324 - Remove Login component's deprecated Message and Error components
- #327 - Externalize USWDS accordion in header to its own component
- #326 - Add sinon spy to test
- Simplify logic in ViewComparisonLink component
- Move display conditions into a constant in PositionDetails component
- Removing extra `it`s in test names


src/Constants/SystemMessages.test.js
```diff
@@ -1,27 +1,19 @@
...
-    expect(SystemMessages.NO_DANGER_PAY).toBeDefined();
-  });
-
-  it('Should return NO_COLA', () => {
-    expect(SystemMessages.NO_COLA).toBeDefined();
-  });
-
-  it('Should return NO_TOUR_OF_DUTY', () => {
-    expect(SystemMessages.NO_TOUR_OF_DUTY).toBeDefined();
+  it('should have all expected messages defined', () => {
```
burgwyn: Did you miss NO_REST_RELAXATION?

src/Components/Header/GovBanner/GovBanner.jsx
```diff
@@ -0,0 +1,55 @@
+import React from 'react';
+import flag from '../../../../node_modules/uswds/dist/img/favicons/favicon-57.png'; // usa flag
```
jseppi: would something like this work:

```js
import flag from 'uswds/dist/img/favicons/favicon-57.png';
```

instead of the big relative path?

src/Constants/SystemMessages.test.js
mjoyce91: 🕵🏻

mjoyce91:
Thanks @burgwyn & @jseppi - see updates.

PR #362, opened 2017-08-14
Make updates to data structure based on API updates

mjoyce91:
This PR provides updates to the expected data structure from the API based on https://github.com/18F/State-TalentMAP-API/pull/81


mjoyce91:
Also added fix for updates found in https://github.com/18F/State-TalentMAP-API/pull/82

PR #386, opened 2017-08-18
Use a div instead of a form to wrap the SelectForm

mjoyce91:
WCAG2AA enforces that `<form>`s have a submit button. Because the SelectForm performs changes on option selection and not on form submission, the following error was reported by pa11y

```
This form does not contain a submit button, which creates issues for those who cannot submit the form using the keyboard. Submit buttons are INPUT elements with type attribute \"submit\" or \"image\", or BUTTON elements with type \"submit\" or omitted/invalid.
```

This PR changes the SelectForm wrapper from a `form` to a `div`.

### DIFF
--- a/src/Components/SelectForm/SelectForm.jsx
+++ b/src/Components/SelectForm/SelectForm.jsx
@@ -29,7 +29,7 @@ class SelectForm extends Component {
       ),
     );
     return (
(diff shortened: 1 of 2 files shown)


burgwyn:
@mjoyce91 - can you update a snapshot?

PR #387, opened 2017-08-18
Add more thorough testing for TotalResults output possibilities

mjoyce91:
Tests against every possible display of `TotalResults` when there are 103 results. Also removes the `&nbsp;`s from the component since they fail against checks of text that uses the `space` key.

### DIFF
--- a/src/Components/TotalResults/__snapshots__/TotalResults.test.jsx.snap
+++ b/src/Components/TotalResults/__snapshots__/TotalResults.test.jsx.snap
@@ -1,15 +1,85 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`TotalResults displays pager numbers and total results correctly when on the first page of the results 1`] = `
+<span
+  id="total-results"
+>
+  Viewing 
+  1
+   - 
+  25
+   of 
+  103
+   results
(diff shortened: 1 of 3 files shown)


src/Components/TotalResults/__snapshots__/TotalResults.test.jsx.snap
```diff
@@ -1,15 +1,85 @@
...
+
 exports[`TotalResults matches snapshot 1`] = `
 <span
   id="total-results"
 >
-  Viewing 
+  Viewing 
   1
    - 
   0
```
burgwyn: I think we should define a `defaultProp` value for `pageSize` and drop the `isRequired`

mjoyce91: Hmm you don't think a `defaultProp` here would be kind of arbitrary? The component needs to know those three props in order to display the right text. `pageSize` should always match the page size of the results being displayed.

burgwyn: Good point.  I'm just looking at ' Viewing 1 - 0 of 103' and it seems weird.

Issue #365, opened 2017-08-15
User profile not retrieved until page refresh


mjoyce91:
Upon logging in, the user's profile data is not loaded into the `AccountDropdown` component. This is due the Header component rendering before the user's profile data is retrieved, resulting in an empty object being passed to `AccountDropdown`. Because the `isAuthorized()` check only occurs on `componentWillMount()`, the user profile should be loaded at another point besides the initial render of the Header. It can be included in the login Saga so that the profile data is retrieved again upon successful login.

Issue #334, opened 2017-08-08
Use "count" instead of "len" as property name to represent length


mjoyce91:
The property name `len` could be confused as a typo for `length`, so use a different name for the property, such as `count`.

Issue #333, opened 2017-08-08
Externalize search result items into their own components


mjoyce91:
Make search results cards their own components.

Issue #332, opened 2017-08-08
Configure any props that control functionality to be required


mjoyce91:
Any props that control functionality, such as `onNavigateTo` or `goBack` should be required. Tests will also need to be updated to provide a function as a prop.

PR #388, opened 2017-08-18
Sprint 5

mjoyce91:
This PR introduces features from Sprint 5, including:

#255 - Sort method for position results
#261 - Display the number of total search results
#299 - Display page size in search results
#300 - Change search results page size

Additional features/fixes include:
#332 - configuring props that control functionality to be `required` 
#333 - externalize search results cards into their own component
#334 - use `count` instead of `len`
#365 - user profile not retrieved until page refresh
#373 - pre-populated login credentials for general public to use


src/Components/ResultsList/ResultsList.test.jsx
```diff
@@ -8,34 +9,37 @@ describe('ResultsListComponent', () => {
...
-      skill: 'OFFICE MANAGEMENT (9017)',
-      bureau: '150000',
-      organization: 'YAOUNDE CAMEROON (YAOUNDE)',
-      position_number: '00025003',
-      is_overseas: true,
-      create_date: '2006-09-20',
-      update_date: '2017-06-08',
-      post: { id: 162, tour_of_duty: '2YRR', code: 'LT6000000', location: 'MASERU, LESOTHO', cost_of_living_adjustment: 0, differential_rate: 15, danger_pay: 0, rest_relaxation_point: 'London', has_consumable_allowance: false, has_service_needs_differential: false },
-      languages: [
+  const resultsArray = {
```
burgwyn: Let's move this into `__mocks__`

mjoyce91: https://github.com/18F/State-TalentMAP/pull/390

mjoyce91:
@burgwyn - it would be nice to have full fledged bread crumbs integrated using something like [this](https://github.com/svenanders/react-breadcrumbs), but I wanted something functional in the mean time, which is why I'm using the history's `goBack` for now.

Issue #300, opened 2017-08-02
Change Search Results Page Size


burgwyn:
As a bidder, 
I want to change the quantity of my search results, 
so that I control the quantity of results visible on the search results page

## Acceptance criteria
* A mechanism to change the quantity of visible search results per page
* Update the page to reflect this change in quantity
* Update the descriptive values regarding results per page

burgwyn:
Closing

Issue #299, opened 2017-08-02
Display Page Size in Search Results


burgwyn:
As a bidder, 
I want to see only a portion of my search results at a time, 
so that I am not overwhelmed by the quantity of results

## Acceptance criteria
* A message indicating the number of results displayed in the page

Issue #261, opened 2017-07-26
Display the number of Results


krisdelacruz:
As a bidder,
I am able to display the number of results on the Search Results page
So that I am able to see more results

Issue #255, opened 2017-07-26
Sort method for position results


wongperson:
### Story
As a user,
I would like to be able to select a sort method for position results
So I can view the available positions in my search criteria organized in a certain order.

### Acceptance criteria
* Acceptance criteria
* Acceptance criteria
* Acceptance criteria [...]

### Definition of done
*

mjoyce91:
@afrasier - is there currently any way to sort by service needs differential (assuming the sort would be all the `true`s followed by all the `false`s. Running a query with `?ordering=post__has_service_needs_differential` doesn't seem to do anything.

PR #392, opened 2017-08-18
Sprint 5 (dev->stg)

mjoyce91:
This PR introduces features from Sprint 5, including:

#255 - Sort method for position results
#261 - Display the number of total search results
#299 - Display page size in search results
#300 - Change search results page size

Additional features/fixes include:
#332 - configuring props that control functionality to be `required` 
#333 - externalize search results cards into their own component
#334 - use `count` instead of `len`
#365 - user profile not retrieved until page refresh
#373 - pre-populated login credentials for general public to use


src/Constants/PropTypes.js
```diff
@@ -70,4 +73,37 @@ export const ITEMS = PropTypes.arrayOf(
...
+  user: PropTypes.shape({
+    username: PropTypes.string,
+    email: PropTypes.string,
+    first_name: PropTypes.string,
+    last_name: PropTypes.string,
+  }),
+  language_qualifications: PropTypes.arrayOf(
+    PropTypes.number,
+  ),
+  favorite_positions: PropTypes.arrayOf(
```
jseppi: Planning to start using this instead of `localStorage` in the next sprint or so?

src/sass/_pagination.scss
```diff
@@ -0,0 +1,30 @@
...
+
+  margin: 15px 0 0;
+
+  ul {
+    display: inline-block;
+    padding-left: 15px;
+    padding-right: 15px;
+  }
+
+  li {
```
jseppi: I think we need some styling for `li.disabled`, which is applied for `preview` and `next` when on the first or last pages of results respectively.

screen shot 2017-08-21 at 11 05 03 am

screen shot 2017-08-21 at 11 04 47 am

src/Containers/Results/Results.jsx
```diff
@@ -2,48 +2,99 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import { connect } from 'react-redux';
 import { push } from 'react-router-redux';
+import queryString from 'query-string';
```
jseppi: Note a huge deal, but I think you could use node's built in [`querystring`](https://nodejs.org/api/querystring.html) instead of having this additional dependency.

src/sass/_pagination.scss
jseppi: Feel free to punt on this until more of a design for the pagination has been done.

src/Components/AccountDropdown/AccountDropdown.jsx
```diff
@@ -13,6 +13,7 @@ export class AccountDropdown extends Component {
   }
 
   render() {
+    const userName = this.props.userProfile.user ? this.props.userProfile.user.username : '...';
     return (
       <Dropdown className="account-dropdown">
         <DropdownTrigger href="/#">
```
mjoyce91: I'll create an issue to work with the design team to determine how this should look in the mobile nav, and then adjust accordingly.

src/Constants/PropTypes.js
mjoyce91: Should be soon - integrating favorites with the backend hasn't been included in the past few sprints.

src/Containers/Results/Results.jsx
mjoyce91: Even if this is all performed client-side?

src/sass/_pagination.scss
mjoyce91: Good idea, I'll add some basic styles to the `disabled` class for now. From talking to @krisdelacruz, it sounds like what we have is pretty close to what he envisioned.

src/Containers/Results/Results.jsx
jseppi: Yup!

PR #393, opened 2017-08-18
Use redux and the router to handle passing history between routes

mjoyce91:
This makes use of `@@router/LOCATION_CHANGE` and redux to store the entire history of the session in an array. This can then be accessed by the `Position` container, and with the help of some mapping, be passed to the `PositionTitle` as a dynamic element such as `<Link to="/results?q=german">Go back to search results</Link>`.

PR'ing into `sprint-6` since this is a bigger change.

### DIFF
--- a/src/Components/PositionTitle/PositionTitle.jsx
+++ b/src/Components/PositionTitle/PositionTitle.jsx
@@ -4,30 +4,20 @@ import PropTypes from 'prop-types';
 import { POSITION_DETAILS } from '../../Constants/PropTypes';
 
 class PositionTitle extends Component { // eslint-disable-line
-  navBack(e) {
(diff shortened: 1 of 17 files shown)


src/Components/PositionTitle/PositionTitle.jsx
```diff
@@ -4,30 +4,20 @@ import PropTypes from 'prop-types';
 import { POSITION_DETAILS } from '../../Constants/PropTypes';
 
 class PositionTitle extends Component { // eslint-disable-line
```
burgwyn: Do we want this?

mjoyce91: Updated

PR #400, opened 2017-08-21
[WIP] - CircleCI 2

burgwyn:


jesse-russell:
@burgwyn in the Dockerfile, I believe we need to add the rest of the code to the image. We can add the line `COPY . /app` to the end.

burgwyn:
@jesse-russell I've updated the Docker file.  Everything runs nicely, but the test results and code coverage are hidden away.  I'll dig a little more.

burgwyn:
@jesse-russell thanks for the help.  

@jseppi do you have any questions?

.circleci/config.yml
```diff
@@ -0,0 +1,171 @@
+version: 2.0
+default-docker-node: &default-docker-node
+  docker:
+    - image: circleci/node:6.11.0
+      environment:
+        CC_TEST_REPORTER_ID: c05c8cb360b66d7b92767ad0ffdeef2c8519f390d700b8bc2d2b7ee6c2760791
+        COMPOSE_FILE_NAME: circleci-docker-compose.yml
+        API_REPO_NAME: state-talentmap-api
+        API_REPO_BRANCH: circleci-2
```
jseppi: Will this need to be changed to `dev`?

jesse-russell: So we can handle that in a couple ways: 
1.  We assume that the UI is always a sprint behind the API. This means we can use the most stable branch of the API - staging.
2.  Or yes, we can use the dev branch, but that means that the build could conflict if breaking changes are made at any point.

Both are valid and both would work. Mostly depends on how we want to handle it.

jseppi: Ah, right, so `staging` seems like the right choice probably moving forward, though the API's staging branch doesn't yet have its CircleCI 2.0 changes incorporated. Want to open a PR for that (https://github.com/18F/State-TalentMAP-API/compare/staging...dev?expand=1)?

jesse-russell: PR is a good idea. I've opened one [here](https://github.com/18F/State-TalentMAP-API/pull/94) and asked you (@jseppi) and @afrasier to approve.

PR #402, opened 2017-08-21
Externalize search bar

mjoyce91:
Externalizes the search bar currently in the `Filters` component into its own `SearchBar` component.

### DIFF
--- a/src/Components/SearchBar/SearchBar.test.jsx
+++ b/src/Components/SearchBar/SearchBar.test.jsx
@@ -3,69 +3,71 @@ import React from 'react';
 import toJSON from 'enzyme-to-json';
 import SearchBar from './SearchBar';
 
-describe('SearchBarComponent', () => {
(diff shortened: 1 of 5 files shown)


src/Components/SearchBar/SearchBar.test.jsx
```diff
@@ -0,0 +1,71 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+import toJSON from 'enzyme-to-json';
+import SearchBar from './SearchBar';
+
+describe('SearchBarComponent', () => {
```
burgwyn: Let's do a describe for 'big' and another for 'small'

burgwyn: And snapshot both

mjoyce91: Updated

PR #405, opened 2017-08-21
Sprint 5 fixes

mjoyce91:
Fixes some line spacing and test name issues found in #392 review.


src/actions/routerLocations.test.jsx
```diff
@@ -0,0 +1,26 @@
+import * as actions from './routerLocations';
```
burgwyn: Rename to `.js`

src/actions/routerLocations.jsx
```diff
@@ -0,0 +1,47 @@
+import React from 'react';
```
burgwyn: Rename to `.js`

mjoyce91: The `getLastRouteLink` function here actually returns a `<Link>`

burgwyn: You're right.  I'm not sure I like an action returning a component.

mjoyce91: I wanted to be flexible for now, in case we decide to just spit out the text and let the component it's eventually passed to decide what to do with the text. But I don't see the text being used as anything _other_ than a link.

burgwyn: What about an object?  { text: '', link: '' } and render the component further downstream?

mjoyce91: Ah yeah, forgot that you'd need to pass the actual link too. The same idea still applies, but I agree, an object is probably a better way to keep presentation-layer stuff out of redux.

Something I noticed that Amazon does is that while they use similar text as us, they just hook into `history.back()`. That UX feels smoother, as to get back to the details page, you'd click the browser's forward button.

mjoyce91: Updated to return object and hook into `window.history.back()`.

src/actions/routerLocations.jsx
```diff
@@ -40,8 +37,8 @@ export function getLastRouteLink(routerLocations) {
   const text = getLastRouteName(routerLocations);
   const lastLocation = getLastRoute(routerLocations);
   if (!text || !lastLocation) {
-    return false;
+    return {};
   }
   const navigateTo = `${lastLocation.pathname}${lastLocation.search}`;
-  return (<Link to={navigateTo}>{text}</Link>);
+  return { text, link: navigateTo };
```
burgwyn: Let's change the file extension too

PR #417, opened 2017-08-22
Responsive account links

mjoyce91:
Makes the `AccountDropdown` only visible if the mobile nav is not enabled. Once the width is small enough for the mobile nav to appear, the `AccountDropdown` is not visible, and its links are included in the mobile nav menu.

### DIFF
--- a/src/Components/Header/Header.jsx
+++ b/src/Components/Header/Header.jsx
@@ -31,7 +31,7 @@ export class Header extends Component {
       },
     } = this.props;
 
-    let showLogin = (<Link to="login">Login</Link>);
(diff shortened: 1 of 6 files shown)


src/Components/Header/Header.jsx
```diff
@@ -65,6 +70,12 @@ export class Header extends Component {
                 </div>
               </form>
               <ul className="usa-unstyled-list usa-nav-secondary-links">
+                <span className="usa-unstyled-list mobile-nav-only">
```
burgwyn: Can you restructure with ul > li > span?

mjoyce91: Restructured slightly differently to override some USWDS styles for `<li>`s

PR #435, opened 2017-08-23
Pagination style

mjoyce91:
Adds style to pagination based on wireframe.

### DIFF
--- a/src/Components/PaginationWrapper/PaginationWrapper.jsx
+++ b/src/Components/PaginationWrapper/PaginationWrapper.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import ReactPaginate from 'react-paginate';
-import FontAwesome from 'react-fontawesome';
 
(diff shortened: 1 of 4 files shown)


src/Components/PaginationWrapper/PaginationWrapper.jsx
```diff
@@ -37,8 +37,8 @@ const PaginationWrapper = ({
   );
 
 PaginationWrapper.propTypes = {
-  previousLabel: PropTypes.string,
-  nextLabel: PropTypes.string,
+  previousLabel: PropTypes.node,
```
burgwyn: Does this need to be `node`?  Do we populate anywhere beyond `defaultProps`?

mjoyce91: I was originally using text and a Font Awesome favicon as the default prop. That still seems like a valid use case, so I left the prop type as `node`. Say for example, we re-used pagination for a different search page, and wanted to pass in an element for the next/previous labels.

PR #439, opened 2017-08-25
Search results cards redesign

mjoyce91:
Adds style for the search results cards and search results sorting. Also adds style to the sort/page size drop-downs.

Makes use of new, separate `Favorite` and `CompareCheck` components for favoriting and sharing, with the intention that the original `FavoritesButton` will be removed once full redesign of site is complete.

Adds a black placeholder div for where filters will eventually be added.

Some responsiveness was added, but focus is to match desktop wireframe.


mjoyce91:
I didn't test pagination after breaking things down into more components. Currently not working, I'll need to fix it.

mjoyce91:
Fixed. I had changed some prop names for simplicity, and forgot to update one.

PR #451, opened 2017-08-29
Pill filter list

mjoyce91:
This PR adds container to Results page with list of pills representing the currently selected filters. When clicked on, the pill and related filter are removed and the page refreshes.

Relies on #404, #439, #445, #449.

### DIFF
--- a/src/Components/PillList/PillList.jsx
+++ b/src/Components/PillList/PillList.jsx
@@ -1,0 +1,28 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import Pill from '../Pill/Pill';
+
+const ResultsContainer = ({ items }) => (
+  <div>
+    {
+      items.map(item =>
+        <Pill name={item.name} code={item.code} onClick={e => this.props.onClick(e)} />,
+      )
+    }
+  </div>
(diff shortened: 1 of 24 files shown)


src/Components/PillList/PillList.jsx
```diff
@@ -0,0 +1,42 @@
...
+import PropTypes from 'prop-types';
+import { PILL_ITEM_ARRAY } from '../../Constants/PropTypes';
+import Pill from '../Pill/Pill';
+
+const shortid = require('shortid');
+
+const PillList = ({ items, onPillClick }) => (
+  <div>
+    {
+      (items.sort((a, b) => { // sort our pills by description
```
burgwyn: Can we pull the sort logic out of `render`?  Maybe push it further up the tree?

jseppi: Might also want to add a `.slice()` before the `.sort(...)` so that the original array isn't mutated.

mjoyce91: Thanks! I resolve this in #456

mjoyce91:
Resolved by #456

Issue #454, opened 2017-09-01
Separate components based on redux state


mjoyce91:
When multiple parts of a page rely on different redux states, wrap each in their own connected component, instead of passing both states to all child components, causing needless re-renders.

https://medium.com/dailyjs/quick-redux-tips-for-connecting-your-react-components-e08da72f5b3

Currently need to implement on Results page, as the filter container only relies on the `filters` state, and the results only rely on the `results` state. When the results refresh, unnecessary re-renders are performed on the on the filter container.

mjoyce91:
This is also why a `setAccordion` function and `selectedAccordion` prop are needed. The accordions on the results page get re-rendered, causing the accordion to collapse. Saving the state to redux allows us to pass it back in. But, this is not ideal.

mjoyce91:
Further updates will need to be made to how the components render. Currently, they are rendered via function, ex: `component={() => (<Home {...props} />)}` opposed to `component={Home}`. This is required for the component to receive the `isAuthorized` function and decide whether or not to load. However, when loaded via function, the entire component is re-rendered when the query params change. Need to consider moving `isAuthorized` some where else, such as a redux state, in order to bypass the need to render components via function.

Issue #257, opened 2017-07-26
Execute application of design to the search results view


mjoyce91:
As a bidder,
I want the search results view to be to have a cleaner look and feel,
So that I can interact with the search results

## Acceptance criteria
- Visual design of the search results view wireframes is applied to the front-end web application

burgwyn:
Closing

PR #456, opened 2017-09-01
Sprint 6 review

mjoyce91:
Adds pagination, filters, search bars, and redesigned search results cards to the results page based on #257.

This is probably the biggest lift of any of our PRs. There's quite a bit of logic involved in mapping the query params to retrieved filters in order to populate filter "pills", and support auto-population of filter checkboxes upon load. I've done my best to comment what is going on.

This also adds some new base components, such as `Accordion` and `AccordionItem`, as well as `CheckBox` and `FieldSet`.

A few notes:
- Need to improve architecture as described in #454 to avoid unnecessary re-renders of child components.
- Some aspects are presentation only, such as the search bar within the filter container, the location search bar, and the language proficiency drop-downs (#445).

In lieu of @burgwyn 's normal review, I've branched the most up-to-date work to `sprint-6-review` for review into `dev` by @jseppi. If everything looks good, no further review will be needed for dev->stg.


mjoyce91:
Codeclimate errors appear to be false positives according to https://github.com/evcohen/eslint-plugin-jsx-a11y/issues/272

@burgwyn - may need to update our version of `eslint-plugin-jsx-a11y`?

jseppi:
@mjoyce91 - it might be the version of `eslint-plugin-jsx-a11y` that's in Code Climate. That service doesn't actually use the version in the project's `package.json`, but rather its own internal version. Looks like they are on `4.0.0`: https://github.com/codeclimate/codeclimate-eslint/blob/e1f1bf63166c4dfdda8c872037d806eb61d9cf05/package.json#L50

I think you should just add a comment and disable eslint for those lines, maybe like this:

```js
// At the time of writing, CodeClimate's version of eslint-a11y-plugin
// did not take role="button" into account with the following error:
// eslint-disable-next-line no-static-element-interactions
```

There is a greenkeeper PR open on CodeClimate to update their plugin: https://github.com/codeclimate/codeclimate-eslint/pull/297

jseppi:
blurggguhgu that didn't work :\

mjoyce91:
@jseppi I think we just need the rule's library reference:

`eslint-disable-next-line jsx-a11y/no-static-element-interactions`

Will push again

jseppi:
Nice, good thinking!

src/Components/ResetFilters/ResetFilters.jsx
```diff
@@ -14,27 +13,32 @@ class ResetFilters extends Component {
     if (!this.state.confirm) {
       this.setState({ confirm: true });
     } else {
-      this.props.onToggle('/');
+      this.props.resetFilters();
```
jseppi: This is a cool interaction! Probably should have a way to answer "No" though, right? :)

src/Components/ResultsFilterContainer/ResultsFilterContainer.jsx
```diff
@@ -0,0 +1,62 @@
...
+          <div className="filter-control-right">
+            <ResetFilters resetFilters={resetFilters} />
+          </div>
+        </div>
+        <br />
+        <div className="usa-grid-full search-filters-container">
+          <SearchFiltersContainer
+            filters={filters}
+            queryParamUpdate={(e) => {
+              onQueryParamUpdate(e); onChildToggle();
```
jseppi: put expressions on separate lines

src/Components/ResultsFilterContainer/ResultsFilterContainer.jsx
```diff
@@ -0,0 +1,62 @@
...
+            }}
+            queryParamToggle={(p, v, r) => {
+              onQueryParamToggle(p, v, r); onChildToggle();
```
jseppi: separate lines please

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
```diff
@@ -0,0 +1,110 @@
...
+        </div>
+      </div>
+    );
+  }
+}
+
+SearchFiltersContainer.propTypes = {
+  filters: FILTER_ITEMS_ARRAY.isRequired,
+  queryParamUpdate: PropTypes.func.isRequired,
+  queryParamToggle: PropTypes.func.isRequired,
```
jseppi: Might want to consider adding a delay or debounce to the actual firing of requests. Like if a user were to select several different Grade checkboxes, the app shouldn't fire off a request for each click.

jseppi: I think that change would happen in `resultsFetchData`

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
```diff
@@ -0,0 +1,110 @@
...
+          searchFilter.item.description === 'skill' ||
+          searchFilter.item.description === 'grade' ||
+          searchFilter.item.description === 'tod' ||
+          searchFilter.item.description === 'region'
+        ),
+    );
+
+    multiSelectFilters.forEach((filters, i) => {
+      if (filters.item.description === 'skill') {
+        multiSelectFilters[i].data = multiSelectFilters[i].data.sort(descriptionSort);
```
jseppi: hehe, I think the interaction on the Region filter will need to be changed. It is a suuuuuper long list:

region-too-long

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
```diff
@@ -0,0 +1,110 @@
...
+
+    // get our normal multi-select filters
+    const multiSelectFilters = this.props.filters.filter(
+      searchFilter =>
+        (
+          searchFilter.item.description === 'skill' ||
+          searchFilter.item.description === 'grade' ||
+          searchFilter.item.description === 'tod' ||
+          searchFilter.item.description === 'region'
+        ),
```
jseppi: I'm not sure how long this will stay around since those filters will probably end up getting some custom treatment, but a potentially cleaner implementation of this logic would be:

```js
const multiSelectFilterNames = ['skill', 'grade', 'tod', 'region'];
const multiSelectFilters = this.props.filters.filter(
  s => multiSelectFilterNames.indexOf(s.item.description) > -1);
```

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
```diff
@@ -0,0 +1,110 @@
...
+    );
+
+    multiSelectFilters.forEach((filters, i) => {
+      if (filters.item.description === 'skill') {
+        multiSelectFilters[i].data = multiSelectFilters[i].data.sort(descriptionSort);
+      }
+    });
+
+    // get our language filter, which we'll render differently
+    const languageFilters = this.props.filters.filter(
```
jseppi: Since I think you only ever want the first result form this, use [`.find(...)`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find) instead of `.filter`

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
```diff
@@ -0,0 +1,110 @@
...
+      searchFilter =>
+        (
+          searchFilter.item.description === 'language'
+        ),
+    );
+
+    // make sure we have an object to use in case there were no languages passed down
+    const languageFilter = languageFilters.length ? languageFilters[0] : { item: {} };
+
+    // languageFilters should only have one object, so we simply call languageFilters[0]
```
jseppi: Looks like this is a slightly misplaced comment -- but it could go away if my suggestion to use `.find` is taken

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
```diff
@@ -0,0 +1,110 @@
...
+        id: `accordion-${languageFilter.item.title}`,
+        expanded: languageFilter.item.title === this.props.selectedAccordion.main,
+      };
+
+    const multiSelectFilterList = multiSelectFilters.map(item => (
+      { content:
+        (<MultiSelectFilter
+          key={item.item.title}
+          item={item}
+          queryParamToggle={(v, p, r) => { this.props.queryParamToggle(v, p, r); }}
```
jseppi: arg names would be great. I also noticed this is in a different order from other places that have `p, v, r`.

jseppi: actually instead of all instances of this pattern 

```jsx
someProp={(someArgs) => { this.props.someFunc(someArgs); }}
```

could you instead just do

```jsx
someProp={this.props.someFunc}
```

e.g., doing this for the highlighted line:

```jsx
queryParamToggle={this.props.queryParamToggle}
```

.circleci/config.yml
```diff
@@ -0,0 +1,171 @@
...
+        COMPOSE_FILE_NAME: circleci-docker-compose.yml
+        API_REPO_NAME: state-talentmap-api
+        API_REPO_BRANCH: staging
+jobs:
+  build:
+    <<: *default-docker-node
+    steps:
+      - checkout
+      - restore_cache:
+          key: dependency-cache-{{ checksum "package.json" }}
```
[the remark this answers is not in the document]
jseppi: ^ @jesse-russell

src/Components/Favorite/Favorite.jsx
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  render() {
+    const text = this.getSavedState() ? 'Remove' : 'Favorite';
+    const iconClass = this.getSavedState() ? 'heart' : 'heart-o';
+    return (
+      // At the time of writing, CodeClimate's version of eslint-a11y-plugin
+      // did not take role="button" into account with the following error:
+      // eslint-disable-next-line jsx-a11y/no-static-element-interactions
+      <div tabIndex="0" role="button" style={{ cursor: 'pointer' }} onClick={() => this.toggleSaved()}>
```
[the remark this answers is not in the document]
jseppi: This is my biggest concern currently -- I think we're going to need some tuning done on this page to make it snappy.

jseppi:
Any idea why `ci/circleci` is hanging?

Oh, is it because that one check has been superseded by all the new split-out `ci/circleci: whatev` checks?

src/Components/Favorite/Favorite.jsx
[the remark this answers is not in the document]
mjoyce91: Okay, looks like the issue was an `onToggle` function I'm using. This is required for the comparisons, because we need to update the text from "Compare" to "Limit reached" when positions are added to the comparison list and the limit is reached, which requires a re-render of the cards. However, this isn't needed for favorites because there's no limit and they can render independent of each other.

src/utilities.js
```diff
@@ -51,4 +51,35 @@ export function fetchUserToken() {
...
+  const A = (a.description || a.code).toLowerCase();
+  const B = (b.description || b.code).toLowerCase();
+  if (A < B) { // sort string ascending
+    return -1;
+  }
+  if (A > B) { return 1; }
+  return 0; // default return value (no sorting)
+};
+
 export default ajax;
```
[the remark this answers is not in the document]
mjoyce91: Good call, ajax is all contained within redux now.

src/Components/PositionDetails/PositionDetails.jsx
```diff
@@ -1,25 +1,25 @@
...
-const PositionDetails = ({ details, api, isLoading, hasErrored, goBack }) => {
+const PositionDetails = ({ details, isLoading, hasErrored, goBackLink }) => {
   const isReady = details && !isLoading && !hasErrored;
   return (
     <div>
       { isReady &&
       <div>
-        <PositionTitle details={details} goBack={goBack} />
+        <PositionTitle details={details} goBackLink={goBackLink} />
         <PositionDetailsItem details={details} />
```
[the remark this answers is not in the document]
mjoyce91: I've had this issue locally as well, but it appears to work on the deployed version. @afrasier

src/Components/Pill/Pill.jsx
```diff
@@ -0,0 +1,25 @@
...
+import FontAwesome from 'react-fontawesome';
+
+const Pill = ({ description, codeRef, selectionRef, onPillClick }) => {
+  const style = {
+    borderRadius: '12px',
+    fontSize: '.8em',
+    padding: '.5em',
+  };
+  return (
+    <button style={style} onClick={() => onPillClick(selectionRef, codeRef, true)}>
```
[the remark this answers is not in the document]
mjoyce91: Good call, that is probably an accessibility concern.

src/Components/Favorite/Favorite.test.jsx
```diff
@@ -0,0 +1,48 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+import TestUtils from 'react-dom/test-utils';
+import Favorite from './Favorite';
+
+describe('Favorite', () => {
+  let favorite = null;
+
+  beforeEach(() => {
+    favorite = TestUtils.renderIntoDocument(<Favorite refKey="0036" type="fav" />);
```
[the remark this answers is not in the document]
mjoyce91: I initially configured it like this to be flexible. For example, if we wanted to use this same component to save a user's "Favorite Posts", we could pass in a different `type` (ex: `favposts`). However, as requirements become more defined, this may not be needed and we can hard-code it.

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
mjoyce91: Maybe some alphabetical sorting as well

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
mjoyce91: Good idea, then I won't need to use `[0]` when I pass it.

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
mjoyce91: I think I meant "array of one object"

src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.jsx
mjoyce91: Good idea, I had considered adding debounce and ajax cancel tokens but didn't have time this sprint.

src/Components/Favorite/Favorite.jsx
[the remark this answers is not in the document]
mjoyce91: The sluggishness has a lot to do with #454 and is what I've been working on yesterday and today to restructure.

.circleci/config.yml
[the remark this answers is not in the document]
jesse-russell: Reviewing the `yarn.lock` file, I agree with you. I think that does make more sense. I'll update that now.

jesse-russell:
Yea I believe that the `ci/circleci` build is the only version.

src/Components/ResetFilters/ResetFilters.jsx
mjoyce91: Maybe just clicking on anything else assumes "No" and it goes back to "Reset Filters"?

mjoyce91: Probably a question for @krisdelacruz

jseppi: yeah, for sure

jseppi:
@jesse-russell: I updated the repo settings to remove the old `ci/circleci` required status check and made all the new ones required instead.

Dockerfile
```diff
@@ -5,3 +5,5 @@ WORKDIR /app
 
 ADD package.json /app/
 RUN yarn
+
+COPY . /app
```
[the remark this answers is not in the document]
jseppi: Perhaps a safer approach to COPYing or ADDing the entire build context directory would be to customize those commands a little more to _only_ include what absolutely needs to be in the images. For example:

```
COPY config /app/config/
COPY public /app/public/
# etc
```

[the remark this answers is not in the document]
jseppi: There's nothing sensitive in the published images currently, but we want to prevent any future potential issues for sure.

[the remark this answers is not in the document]
jseppi: Filed https://github.com/18F/State-TalentMAP/issues/469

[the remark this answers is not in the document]
jseppi: and https://github.com/18F/State-TalentMAP-API/issues/113

src/Components/SearchFilters/BooleanFilter/BooleanFilter.jsx
```diff
@@ -0,0 +1,34 @@
...
+
+const BooleanFilter = ({ item, onBooleanFilterClick }) => (
+  <div>
+    <FieldSet key={item.item.title} legend={item.item.title}>
+      <CheckBox
+        id={`checkbox-${item.item.title}`}
+        label="Yes"
+        title={item.item.title}
+        name={item.item.title}
+        value={item.data[0].isSelected || false}
```
jseppi: The `item.data[0]` stuff is a little strange to me in the case of these boolean filters. I can't think of a good suggestion for how to fix right now though...maybe instead of `FILTER_ITEM`, this should take a `BOOLEAN_FILTER_ITEM` that is a simplified form of `FILTER_ITEM`.

src/Components/SearchFilters/MultiSelectFilter/MultiSelectFilter.jsx
```diff
@@ -0,0 +1,40 @@
...
+import { FILTER_ITEM } from '../../../Constants/PropTypes';
+
+const MultiSelectFilter = ({ item, queryParamToggle }) => (
+  <FieldSet key={item.item.title} legend={item.item.title}>
+    {
+      item.data.map(itemData => (
+        <CheckBox
+          id={`checkbox${itemData.long_description || itemData.description || itemData.code}`}
+          key={`${item.item.selectionRef}-${itemData.code}`}
+          label={itemData.long_description || itemData.description || itemData.code}
```
jseppi: maybe extract this logic to a reusable function and use it on this line and the next two:

```js
const getItemLabel = (itemData) => itemData.long_description || itemData.description || itemData.code;
```

Dockerfile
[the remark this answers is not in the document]
jesse-russell: Created a PR for this repo #483

jseppi:
The build failed on Circle, but it looks like it might be due to a (hopefully temporary) `yarn` outage. I will restart it.

jseppi:
Build fixed - approved ✅

src/Components/ResultsPage/ResultsPage.jsx
```diff
@@ -23,103 +21,59 @@ class Results extends Component {
...
+            pageCount={pageCount}
+            hasLoaded={hasLoaded || false}
+            defaultSort={defaultSort}
+            pageSizes={pageSizes}
+            defaultPageSize={defaultPageSize}
+            defaultPageNumber={defaultPageNumber}
+            queryParamUpdate={onQueryParamUpdate}
+            refreshKey={this.state.key}
+            onToggle={() => this.onChildToggle()}
+            pillFilters={pillFilters}
```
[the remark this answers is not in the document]
mjoyce91: When you say state, do you mean the `PillList` component's state? This one is tricky, because new pills pop up if a filter is selected from the left-hand filter container, and I'd like the redux state to be the single source of truth.

jseppi:
Shall we go ahead and merge this? :shipit:

PR #493, opened 2017-09-07
Add keyword search redesign to homepage

mjoyce91:
Adds the redesigned search bar to the homepage, which accepts keyword searches and upon submission, directs the user to the results page with that keyword. Makes use of the same `ResultsSearchHeader` component as the Results page for consistency (though some styles differ between the home and results pages). Clears out old filters to prep for additional redesign tasks.


src/Containers/Home/Home.jsx
```diff
@@ -50,7 +49,7 @@ class Home extends Component {
 Home.propTypes = {
   onNavigateTo: PropTypes.func.isRequired,
   fetchData: PropTypes.func,
-  isLoading: PropTypes.bool,
+  // isLoading: PropTypes.bool,
```
burgwyn: Remove

src/sass/_resultsSearchBarContainer.scss
```diff
@@ -45,25 +46,22 @@
     label {
       display: block;
       text-align: left;
+      text-transform: uppercase;
     }
   }
 
-  .search-location {
-    padding: 0 30px;
+  .usa-button {
```
burgwyn: Are we overriding a USWDS style here?  Should we use a `tm-` instance instead?

src/sass/_variables.scss
```diff
@@ -7,6 +7,7 @@ $blue-primary: #0071bc;
 $color-white: #fff;
 $color-black: #000;
 $color-gray: #5b616b;
+$tm-dark-blue: #00293f;
```
burgwyn: Is the thinking that we'll identify our custom colors as `tm-`?

src/Components/Homepage/Homepage.jsx
```diff
@@ -0,0 +1,48 @@
+import React, { Component } from 'react';
```
burgwyn: Let's move this from Components to Containers

src/sass/_resultsSearchBarContainer.scss
mjoyce91: This is only within the context of the `results-search-bar-container` class.

src/sass/_variables.scss
mjoyce91: Right, opposed to the generic USWDS colors.

PR #503, opened 2017-09-07
Explore by Region section

mjoyce91:
Adds in functionality for #474 and #475 by adding an Explore by Region section to the homepage, which includes a dropdown of regions and a static world map image. Users can select a region and click to view results of that region.

Also externalizes some of the query params that are used to filter the results.


mjoyce91:
@burgwyn have we updated anything with pa11y on circleci? Seems like we're getting a false positive.

mjoyce91:
🎉 For some reason `.find` was breaking pa11y, so I instead used `.filter` and `[0]`. @burgwyn or @jseppi - are either of you aware of any cross-browser compatibility issues with `find`?

burgwyn:
Looks like find and findIndex are unavailable in IE https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

package.json
```diff
@@ -22,6 +22,7 @@
     "redux-saga": "^0.15.4",
     "redux-thunk": "^2.2.0",
     "shortid": "^2.2.8",
+    "typeface-vollkorn": "0.0.35",
```
burgwyn: @shanzasubz - are we working with Volkhorn?

src/Components/ExploreRegionDropdown/ExploreRegionDropdown.jsx
```diff
@@ -0,0 +1,94 @@
...
+
+  searchRegion(e) {
+    e.preventDefault();
+    this.props.onRegionSubmit(this.state.selection.value);
+  }
+
+  render() {
+    const { filters } = this.props;
+    // function to find the Region filters
+    function filterRegion(filterItem) {
```
burgwyn: Let's pull all this logic out of `render` and into a private method

src/Components/FieldSet/FieldSet.jsx
```diff
@@ -1,16 +1,21 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 
-const FieldSet = ({ legend, children }) => (
+const FieldSet = ({ legend, children, legendSrOnly }) => (
   <fieldset>
-    <legend>{legend}</legend>
+    <legend className={legendSrOnly ? 'usa-sr-only' : null}>{legend}</legend>
```
burgwyn: 👍

src/Constants/EndpointParams.js
```diff
@@ -0,0 +1,13 @@
+const ENDPOINT_PARAMS = {
```
burgwyn: Are these for all endpoints or just positions?

src/sass/_variables.scss
```diff
@@ -8,6 +8,7 @@ $color-white: #fff;
 $color-black: #000;
 $color-gray: #5b616b;
 $tm-dark-blue: #00293f;
+$tm-white-smoke: #f5f5f5;
```
burgwyn: 💨

src/Constants/EndpointParams.js
mjoyce91: Just positions.

package.json
mjoyce91: screen shot 2017-09-08 at 8 47 36 am
screen shot 2017-09-08 at 8 47 43 am

mjoyce91:
Issues with circle, will rebuild later.

https://status.circleci.com/

jseppi:
Hrm, regarding the missing `.find` method in IE. We might want to include Babel's polyfill (https://github.com/babel/babel-preset-env#usebuiltins). What do you all think? If this is the only thing missing that we want, then it's probably not worth it, but if more things come up it might be a good idea.

PR #510, opened 2017-09-11
Homepage Redesign - Position Sections

mjoyce91:
Adds the New and Popular Positions sections to the Homepage.

What this PR covers:
- Presentation layer of new and popular positions on the home page
- Appropriate prop types
- "View More" link

What this PR does not cover:
- Tests
- Passing real data as props to the components (i.e. the actual results from the API)
- Any real action associated with the "Add to Bid List" button
- How to check if a position is considered "new"
- Actual results property names for placeholders
- A URL to use for Popular Positions "View More" link (straight `/results` for now)

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,6 +22,7 @@
     "redux-saga": "^0.15.4",
     "redux-thunk": "^2.2.0",
     "shortid": "^2.2.8",
+    "typeface-roboto": "0.0.35",
     "typeface-vollkorn": "0.0.35",
     "uswds": "^1.3.0"
   },
--- a/package.json
+++ b/package.json
@@ -87,7 +87,8 @@
   "jest": {
     "collectCoverageFrom": [
       "src/**/*.{js,jsx}",
-      "!src/registerServiceWorker.js"
+      "!src/registerServiceWorker.js",
+      "!src/Components/**/index.js"
     ],
     "coverageReporters": [
       "lcov"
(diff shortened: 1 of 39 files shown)


src/Components/PositionsSectionContent/PositionsSectionContent.jsx
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
```
burgwyn: This file is uncovered

src/Components/PositionsSectionContent/index.js
```diff
@@ -0,0 +1 @@
+export { default } from './PositionsSectionContent';
```
burgwyn: Let's add `!src/Components/**/index.js` to our `package.json > jest > collectCoverageFrom` so we don't worry about these files in our coverage

src/Components/PositionsSectionContent/PositionsSectionContent.jsx
mjoyce91: I decided to not do any tests to split things up evenly between two PRs.

src/Components/PositionsSectionContent/index.js
mjoyce91: Good point, I forgot to mention that I introduced this style of importing components. Makes the references shorter (`'../PositionsSectionContent/PositionsSectionContent'` -> `'../PositionsSectionContent'`).

mjoyce91: Updated

burgwyn:
Can you commit `yarn.lock`?

package.json
```diff
@@ -22,6 +22,7 @@
     "redux-saga": "^0.15.4",
     "redux-thunk": "^2.2.0",
     "shortid": "^2.2.8",
+    "typeface-roboto": "0.0.35",
```
burgwyn: Need the updated `yarn.lock`

PR #519, opened 2017-09-12
Homepage Redesign - Position Sections + Redux

mjoyce91:
This PR adds functionality to retrieve and pass down positions from the API to the "Popular" and "New" Positions sections using redux. The ability to favorite positions from the new condensed cards has been added. It also adds test coverage for this section and some minor style fixes.

Note: There is currently no "popular" positions end point, so "Popular Positions" -> "View More >" simply directs to the base `/results` endpoint with no additional query params.


mjoyce91:
CircleCI is working again 🙌🏼

PR #523, opened 2017-09-13
Header and footer style edits

mjoyce91:
Some small edits to the header and footer, including changing the font size of the title in the header, using a non-JS/dynamic search button, and adding color the footer nav.

### DIFF
--- a/src/Components/Header/Header.jsx
+++ b/src/Components/Header/Header.jsx
@@ -66,7 +66,7 @@ export class Header extends Component {
               <img src={close} alt="close" />
             </button>
             <div className="usa-nav-secondary">
-              <form className="usa-search usa-search-small js-search-form usa-sr-only">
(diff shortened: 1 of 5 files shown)


src/Components/Header/Header.jsx
```diff
@@ -66,7 +66,7 @@ export class Header extends Component {
               <img src={close} alt="close" />
             </button>
             <div className="usa-nav-secondary">
-              <form className="usa-search usa-search-small js-search-form usa-sr-only">
+              <form className="usa-search usa-search-small usa-sr-only">
```
burgwyn: Does USWDS require to `js-*` classes?

mjoyce91: If you look at what's currently on the dev site, the Search button opens its own mini search using that class. Depending on style implementation, we'd want to connect that button to the existing search bar.

PR #524, opened 2017-09-13
Accessibility fix for SearchBar component

mjoyce91:
SearchBar component did not have necessary `name` attribute for its buttons. Additionally, multiple SearchBar components could not be used on the same page, as the `id`s were not unique enough. This PR fixes both of those issues.


mjoyce91:
pa11y errors should resolve when #522 is merged.

Issue #490, opened 2017-09-06
Extract item label into a reusable function for the MultiSelectFilter component


mjoyce91:
Example:

```js
const getItemLabel = (itemData) => itemData.long_description || itemData.description || itemData.code;
```

PR #525, opened 2017-09-13
Externalize item label function

mjoyce91:
Addresses #490.


src/utilities.js
```diff
@@ -101,3 +101,7 @@ export const formExploreRegionDropdown = (filters) => {
   }
   return regions;
 };
+
+// when we want to grab a label, but arne't sure which one exists
```
burgwyn: typo in the comment

PR #527, opened 2017-09-13
Scroll to top on pagination change

mjoyce91:
Introduces the `react-scroll` library to scroll to the top of the window when the page number is changed on the Results page. Also adds some `this` bindings to functions on the Results component.

Tried to mimic [USAJobs](https://www.usajobs.gov/Search/?k=technology) while keeping in mind that our site is a little different.

### DIFF
--- a/src/utilities.js
+++ b/src/utilities.js
@@ -1,3 +1,7 @@
+import Scroll from 'react-scroll';
+
+const scroll = Scroll.animateScroll;
+
 export function localStorageFetchValue(key, value) {
   const saved = { exists: true, count: 0 };
   const retrievedKey = localStorage.getItem(key);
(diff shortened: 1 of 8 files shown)


src/utilities.js
```diff
@@ -102,6 +106,14 @@ export const formExploreRegionDropdown = (filters) => {
   return regions;
 };
 
+export const scrollToTop = () => {
```
burgwyn: Do we want to make this function params with these as defaults?

mjoyce91: We can do that. I figured we'd want to keep the transition consistent, but it wouldn't hurt to allow an optional config object.

src/utilities.js
```diff
@@ -102,6 +106,17 @@ export const formExploreRegionDropdown = (filters) => {
...
 };
 
+// see all props at https://github.com/fisshy/react-scroll#propsoptions
+const defaultScrollConfig = {
+  duration: 700,
+  delay: 270,
+  smooth: 'easeOutQuad',
+};
+
+export const scrollToTop = (config = defaultScrollConfig) => {
```
burgwyn: 👍

Issue #464, opened 2017-09-06
Use full parameter names for QueryParamUpdate and QueryParamToggle functions


mjoyce91:
Instead of passing `(p, v, r)`, use unabbreviated param names to make it more clear what the function is passing.

PR #542, opened 2017-09-13
Unabbreviated params

mjoyce91:
This PR updates function calls to use shorthand when possible (`someFunction` instead of `() => someFunction()` by binding the functions to `this`. Also addresses #464 by using unabbreviated parameter names when possible.


jseppi:
Potentially useful: https://www.npmjs.com/package/react-autobind

PR #553, opened 2017-09-14
Capsule descriptions

mjoyce91:
Adds real capsule descriptions from the API to the Results cards and Position Details page. Adds a new `shortenString` function to shorten long capsule descriptions and add an ellipsis to the end.

### DIFF
--- a/src/utilities.js
+++ b/src/utilities.js
@@ -105,3 +105,15 @@ export const formExploreRegionDropdown = (filters) => {
 // when we want to grab a label, but aren't sure which one exists
 export const getItemLabel = itemData =>
   itemData.long_description || itemData.description || itemData.code;
+
+// shortens descriptions to varying lengths
+export const shortenString = (string, shortenBy = 250) => {
+  let newString = string;
+  if (string.length > shortenBy) {
+    newString = string.slice(0, shortenBy); // shorten by shortenBy
+    newString = newString.trim(); // in case the last character(s) was whitespace
+    newString += '...'; // append ellipsis
+    return newString; // return newly formed string
+  }
+  return newString;
+};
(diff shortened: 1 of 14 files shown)


src/utilities.js
```diff
@@ -120,3 +120,15 @@ export const scrollToTop = (config = defaultScrollConfig) => {
 // when we want to grab a label, but aren't sure which one exists
 export const getItemLabel = itemData =>
   itemData.long_description || itemData.description || itemData.code;
+
+// shortens descriptions to varying lengths
+export const shortenString = (string, shortenBy = 250) => {
```
burgwyn: 👍

jseppi: arg name might be a little clearer if it was something like `maxLength`, or `shortenTo`. `shortenBy` gives me the impression that is is subtracting that number of characters, rather than truncating to it.

src/utilities.test.js
```diff
@@ -127,3 +128,10 @@ describe('getItemLabel function', () => {
     expect(getItemLabel({ code: '0' })).toBe('0');
   });
 });
+
+describe('shortenString function', () => {
+  it('can shorten a string', () => {
+    const string = '012345';
+    expect(shortenString(string, 3)).toBe('012...');
```
jseppi: heh, interestingly this test case shows that the the function actually returns a string of the same length and it isn't shortened at all 🤔 maybe it should take into account the length of the ellipsis (ie, `3`) before determining whether or not to do anything.

mjoyce91: Ha, yeah I see what you mean. It's still doing what we want, but I guess `shortenBy` doesn't become an accurate description. I was using `shortenBy` somewhat arbitrarily, opposed to defining it as "I want the string to be this many characters, including the ellipsis".

src/utilities.js
mjoyce91: That's a good point, and I realized that while replying to your other comment.

src/utilities.test.js
burgwyn: `shortenString('12345', 3).ellipsis()`?

PR #558, opened 2017-09-15
API Favoriting

mjoyce91:
Migrates favoriting from local storage to the API/user profile.

### DIFF
--- a/src/Components/FavoritesButton/FavoritesButton.jsx
+++ b/src/Components/FavoritesButton/FavoritesButton.jsx
@@ -1,66 +1,43 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { localStorageFetchValue, localStorageToggleValue } from '../../utilities';
-import { EMPTY_FUNCTION } from '../../Constants/PropTypes';
 
 class FavoritesButton extends Component {
   constructor(props) {
     super(props);
     this.toggleSaved = this.toggleSaved.bind(this);
-    this.state = {
-      saved: false,
-      localStorageKey: null,
-      count: 0,
-    };
-  }
-
-  componentWillMount() {
-    const localStorageKey = this.props.type;
-    this.setState({ localStorageKey });
-  }
-
-  componentDidMount() {
-    this.getSaved();
-  }
-
-  onToggle() {
-    this.props.onToggle();
-  }
-
-  getSaved() {
-    const saved = localStorageFetchValue(this.state.localStorageKey, this.props.refKey);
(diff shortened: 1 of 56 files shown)


mjoyce91:
Don't let the diff scare you - it's mostly the snapshots and required props in the tests.

src/actions/userProfile.js
```diff
@@ -20,16 +20,87 @@ export function userProfileFetchDataSuccess(userProfile) {
...
+export function userProfileToggleFavoritePosition(id, remove) {
+  const idString = id.toString();
+  return (dispatch) => {
+    dispatch(userProfileFavoritePositionIsLoading(true));
+    dispatch(userProfileFavoritePositionHasErrored(false));
+    axios.get(`${api}/profile/`, { headers: { Authorization: fetchUserToken() } })
+            .then(response => response.data)
+            .then((userProfile) => {
+              // the user's refreshed favorites
+              let favorites = userProfile.favorite_positions;
```
burgwyn: Take a look at [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set)

src/utilities.js
```diff
@@ -132,3 +132,14 @@ export const shortenString = (string, shortenBy = 250) => {
   }
   return newString;
 };
+
+// for checking if a favorite_position exists in the user's profile
+export const existsInArray = (ref, array) => {
```
burgwyn: [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) - check out `has`

src/Components/FavoritesButton/FavoritesButton.jsx
```diff
@@ -1,66 +1,39 @@
...
+    const savedState = this.getSavedState();
+    const buttonClass = savedState ? 'usa-button-secondary' : '';
+    const buttonText = savedState ? `Remove from ${fromText}` : `Add to ${fromText}`;
+    const style = {
+      pointerEvents: this.props.isLoading ? 'none' : 'inherit',
+      cursor: 'pointer',
+    };
     return (
-      <div>
+      <div style={style}>
```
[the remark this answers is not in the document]
mjoyce91: I kept these styles in the component because I felt that they were closely tied with its functionality. Otherwise, I keep them in stylesheets.

src/actions/userProfile.js
mjoyce91: I'll open an issue to look into this more. I also have some questions about the best approach to adding and removing favorites via the API.

Issue #554, opened 2017-09-14
Use better parameter word/verbage for shortenString function


mjoyce91:
Make it clear what length the function shortens the original string to, and use parameter names that more clearly define their purpose.

PR #562, opened 2017-09-15
Update verbiage for shortenString function

mjoyce91:
Addresses #554.

### DIFF
--- a/src/utilities.js
+++ b/src/utilities.js
@@ -122,20 +122,23 @@ export const getItemLabel = itemData =>
 
 // abcde 4 // a...
 // Shortens strings to varying lengths
-export const shortenString = (string, shortenTo = 250) => {
+export const shortenString = (string, shortenTo = 250, suffix = '...') => {
   let newString = string;
+  let newSuffix = suffix;
+  if (!newSuffix) {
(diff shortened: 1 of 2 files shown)


burgwyn:
Let's pull the ellipsis out and make it a separate function.  The `shortenBy` loses it's potential utility if we force in the ellipsis.

mjoyce91:
How about something like this? Set a default suffix, but allow passing in a custom one, or `''` or `null` to simply shorten the string without a suffix.

```js
export const shortenString = (string, shortenTo = 250, suffix = '...') => {
  let newString = string;
  let newSuffix = suffix;
  if (!newSuffix) {
    newSuffix = '';
  }
  // return the suffix even if the shortenTo is less than its length
  if (shortenTo < newSuffix.length) {
    return suffix;
  }
  if (string.length > shortenTo) {
    // shorten to the shortenTo param, less the length of our suffix
    newString = string.slice(0, shortenTo - newSuffix.length);
    // in case the last character(s) was whitespace
    newString = newString.trim();
    // append suffix
    newString += newSuffix;
  }
  // return the string
  return newString;
};
```

Issue #559, opened 2017-09-15
Use Set in the userProfile Favoriting Action 


mjoyce91:
Based on feedback from https://github.com/18F/State-TalentMAP/pull/558
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

We should also consider if `PATCH`ing the entire array is the best approach, compared to performing adds/removes at an individual level. The current action performs three requests which might not be ideal (1. Refresh the entire array, 2. add/remove the value, 3. refresh the the user profile).

@afrasier

PR #564, opened 2017-09-18
Use Set for Favoriting

mjoyce91:
Addresses #559 by using `Set` in the favoriting action.

### DIFF
--- a/src/actions/userProfile.js
+++ b/src/actions/userProfile.js
@@ -69,21 +69,19 @@ export function userProfileToggleFavoritePosition(id, remove) {
     axios.get(`${api}/profile/`, { headers: { Authorization: fetchUserToken() } })
             .then(response => response.data)
             .then((userProfile) => {
-              // the user's refreshed favorites
-              let favorites = userProfile.favorite_positions;
-              // map down to an array of IDs
-              favorites = favorites.map(f => f.id.toString());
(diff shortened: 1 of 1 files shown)


src/actions/userProfile.js
```diff
@@ -71,20 +71,16 @@ export function userProfileToggleFavoritePosition(id, remove) {
             .then((userProfile) => {
               // the user's refreshed favorites
               let favorites = userProfile.favorite_positions;
+              // map down to an array of IDs
```
burgwyn: Can we roll to a single line `let favorites = userProfile.favorite_positions.map(f => f.id.toString());`

src/actions/userProfile.js
```diff
@@ -71,20 +71,16 @@ export function userProfileToggleFavoritePosition(id, remove) {
...
               favorites = favorites.map(f => f.id.toString());
-              // the index of the position in question
-              const indexOfFavorite = favorites.indexOf(idString);
+              // convert the array to a Set
+              favorites = new Set(favorites);
```
burgwyn: And create a separate `favoritesSet` object here?

src/actions/userProfile.js
mjoyce91: Makes sense!

Issue #580, opened 2017-09-19
Add opacity class to Home page results cards when loading


mjoyce91:
When the results cards on the Home Page are loading, use the same class as the results page cards on the Results Page to add opacity to the container. Also, the spinner should be in front of the cards.

PR #590, opened 2017-09-19
Add opacity to home page positions when loading

mjoyce91:
Addresses #580 by adding opacity to the containers for New and Popular positions on the home page when they're loading.

### DIFF
--- a/src/sass/_spinner.scss
+++ b/src/sass/_spinner.scss
@@ -5,6 +5,7 @@ $tm-spinner-filled-ratio: 1.15;
 
 .tm-spinner {
   color: $tm-navy;
+  z-index: 100; // make sure spinner is on top
 }
 
 .tm-spinner-homepage-position-results {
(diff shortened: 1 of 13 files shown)


burgwyn:
This seems like the perfect item to document with gifs

mjoyce91:
b7r4vktpyv

src/sass/_spinner.scss
```diff
@@ -5,6 +5,7 @@ $tm-spinner-filled-ratio: 1.15;
 
 .tm-spinner {
   color: $tm-navy;
+  z-index: 100; // make sure spinner is on top
```
burgwyn: Let's extract the `z-index` into a global variable so we can better track all the potential `z-index` values we have/may have.

mjoyce91:
@burgwyn I can see that as a possibility - multiple lists loading on the same page.

Issue #573, opened 2017-09-19
Use Promise.all for multiple requests in one action


mjoyce91:
When multiple API requests take place in one action, such as for the filters and home page positions, use `Promise.all` instead of checking the array length.

PR #592, opened 2017-09-19
Promise.all for returning multiple requests at once

mjoyce91:
Addresses #573 by using Promise.all for returning batches of requests for filters and positions.

### DIFF
--- a/src/actions/filters.js
+++ b/src/actions/filters.js
@@ -103,10 +103,11 @@ export function filtersFetchData(items, queryParams, savedResponses) {
     if (savedResponses) {
       dispatchSuccess();
     } else {
-      dispatch(filtersHasErrored(false));
+      // our static filters
       const staticFilters = items.filters.slice().filter(item => (!item.item.endpoint));
       responses.filters.push(...staticFilters);
 
+      // our dynamic filters
       const dynamicFilters = items.filters.slice().filter(item => (item.item.endpoint));
       const queryProms = dynamicFilters.map(item => (
         axios.get(`${api}/${item.item.endpoint}`)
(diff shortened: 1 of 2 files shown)


src/actions/filters.js
```diff
@@ -35,69 +35,66 @@ export function filtersFetchData(items, queryParams, savedResponses) {
...
+                    if (filterItemObject.code &&
                           filterItemObject.code.toString() === mappedObject.codeRef.toString() &&
                           filterItem.item.selectionRef === mappedObject.selectionRef) {
-                        responses.filters[i].data[ii].isSelected = true;
-                        if ( // boolean filters are special since they don't rely on AJAX
+                      responses.filters[i].data[ii].isSelected = true;
+                      if ( // boolean filters are special since they don't rely on AJAX
                           response.item.title === 'COLA' ||
                           response.item.title === 'Post Differential' ||
                           response.item.title === 'Danger pay' ||
```
burgwyn: Did I change the spelling on this?

mjoyce91:
How do you see Set being used?

src/actions/filters.js
burgwyn: I made an update here in `4-month-demo` https://github.com/18F/State-TalentMAP/blob/4-month-demo/src/reducers/filters.js

mjoyce91: Ah, the capitalization

mjoyce91: I include `description` in the same object. That was actually its purpose, in case we ever decided to change the display name (aka `title`). Not sure why I chose to use `title` here, but I can update it to use `description`.

burgwyn:
I think Set might be overkill on second thought

src/actions/filters.js
burgwyn: Can you switch to `description`?

mjoyce91: Will do today.

mjoyce91: Updated!

PR #645, opened 2017-09-22
Popular -> Highlighted positions

mjoyce91:
Changes "Popular" to "Highlighted" positions. Uses the `/position/highlighted/` endpoint to populate the card list. Currently there is no filter for Highlighted positions on the Results page, so "View More" directs to the base `/results` route to avoid confusion.


burgwyn:
I'd like to merge in `4-month-demo` before reviewing this.  I think there may be a few, minor conflicts.

mjoyce91:
@burgwyn just merged `sprint-8` into this since it also includes `4-month-demo`.

mjoyce91:
@burgwyn I may have misspoke. `sprint-8` does not include `4-month-demo`. However, I think we could address this in a separate PR of 4-month-demo -> sprint-8.

PR #646, opened 2017-09-22
4 month demo

burgwyn:
Minor content edits to close out our 4 month prototype


src/Components/ViewDetailsButton/ViewDetailsButton.jsx
```diff
@@ -0,0 +1,19 @@
...
+import PropTypes from 'prop-types';
+import FontAwesome from 'react-fontawesome';
+import LinkButton from '../LinkButton';
+
+const ViewDetailsButton = ({ id }) => (
+  <LinkButton toLink={`/details/${id}`} className="white-button">
+    <span className="white-button-icon">
+      <FontAwesome name="plus-circle" />
+    </span>
+    View details
```
mjoyce91: Do we want to keep the View details button on the Home Page results? This was a switch we made for the demo, but wasn't in the mockups. I believe we switched to using an `<a>` tag on the position title.

burgwyn: Let's leave it for now.  This is meant to be a snapshot in time.  We can create the issue in `sprint-8` to add the `a`.

Issue #433, opened 2017-08-23
Visualize - Clone an existing saved search


burgwyn:
As a user,
I want to be able to clone an existing saved search
So that I can re-use existing search parameters when making edits for a new search. 

## Acceptance Criteria

- User can clone the list of saved search parameters
- User can rename the search
- User receives confirmation that the edits were saved
- Default name would be previous search + a number, eg "My Search 1"

PR #657, opened 2017-09-25
Edit a Saved Search

mjoyce91:
Addresses #432 by adding functionality to edit existing saved searches. This includes functionality for storing the currently selected saved search in redux state. Existing searches can be re-saved, or "Saved as". The "Save as" button is not meant to address #433, but should provide some functionality that can be re-used when that feature is addressed.

### DIFF
--- a/src/Components/SaveNewSearchDialog/SaveNewSearchDialog.jsx
+++ b/src/Components/SaveNewSearchDialog/SaveNewSearchDialog.jsx
@@ -47,18 +47,18 @@ class SaveNewSearchDialog extends Component {
           />
         </FieldSet>
         <div className="saved-search-form-buttons">
-          <button className="saved-search-form-submit" type="submit">{ currentSearchExists ? 'Save As' : 'Save' }</button>
+          <button className="saved-search-form-primary-button" type="submit">{ currentSearchExists ? 'Save As' : 'Save' }</button>
           {
             currentSavedSearch.id ?
               <button
-                className="saved-search-form-saveas"
+                className="saved-search-form-secondary-button"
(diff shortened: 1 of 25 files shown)


src/sass/_results.scss
```diff
@@ -339,6 +339,11 @@
     padding: 1rem 2rem;
   }
 
+  .saved-search-form-saveas {
```
burgwyn: `save-as`

burgwyn: `save-as`.  Also, can we possibly collapse these into a single class?

src/Components/SaveNewSearchDialog/SaveNewSearchDialog.jsx
```diff
@@ -1,52 +1,83 @@
...
+            value={currentSavedSearch.name}
+            labelMessage={newSavedSearchHasErrored || newSavedSearchSuccess}
+          />
+        </FieldSet>
+        <div className="saved-search-form-buttons">
+          <button className="saved-search-form-submit" type="submit">{ currentSearchExists ? 'Save As' : 'Save' }</button>
+          {
+            currentSavedSearch.id ?
+              <button
+                className="saved-search-form-saveas"
```
burgwyn: We're using `save-as` class on Save

mjoyce91: To avoid confusion, I'll use `primary` and `secondary` as a suffix. I wanted the button that represents saving a new search to always be first, therefore it's text is conditionally either "Save" or "Save As" depending on if a saved search object exists.

PR #680, opened 2017-09-29
Compare buttons

burgwyn:
We got feedback from the Product Owner to swap the location and prominence of the compare and reset buttons.


src/sass/_results.scss
```diff
@@ -151,22 +151,16 @@
 }
 
 .top-nav {
-  .compare-link {
+  .reset-compare-link {
```
mjoyce91: Can we add `cursor: pointer` to `a`?

```css
.reset-compare-link {
    float: left;
    padding: 15px;

    a {
      cursor: pointer;
    }
  }
```

PR #696, opened 2017-10-02
[TEST COMMIT] to see if CodeClimate works

jseppi:
**do not merge**


jseppi:
I was thinking it might be the size of webhook payloads causing CC to choke, but this is a very small PR and the check is still hanging here...

PR #708, opened 2017-10-03
Logo

burgwyn:
Adding logos and updating header


jseppi:
🎉 
favicon coming, as well?

burgwyn:
Totally would have missed that, thanks!

PR #727, opened 2017-10-05
Append abbreviation to Bureau

mjoyce91:
Appends the Bureau code to the Bureau name within the Results filters container.

Initially I opted to run this logic at the component level, but decided to put in the redux action so that it can easily be used for other forms, such as the Region dropdown on the home page. Then  `custom_description` was set as the first `OR` check in `getItemLabel` which encourages some consistency across the site for any custom labels.

### DIFF
--- a/src/actions/filters.js
+++ b/src/actions/filters.js
@@ -38,8 +38,8 @@ export function filtersFetchData(items, queryParams, savedResponses) {
       // Set all of our isSelected values back to false.
       // We'll check if they should be set to true later
       responses.filters.forEach((responseFilter, i) => {
-        responseFilter.data.forEach((responseFilterData, ii) => {
-          responses.filters[i].data[ii].isSelected = false;
+        responseFilter.data.forEach((responseFilterData, j) => {
+          responses.filters[i].data[j].isSelected = false;
         });
       });
       // check for option queryParamObject to map against (used for pill filters)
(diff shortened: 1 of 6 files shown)


src/actions/filters.js
```diff
@@ -44,15 +44,28 @@ export function filtersFetchData(items, queryParams, savedResponses) {
       });
       // check for option queryParamObject to map against (used for pill filters)
       responses.mappedParams = [];
+
+      // set any custom descriptions
+      responses.filters.forEach((filterItem, i) => {
+        filterItem.data.forEach((filterItemObject, ii) => {
```
burgwyn: ii -> j

src/actions/filters.js
```diff
@@ -44,15 +44,28 @@ export function filtersFetchData(items, queryParams, savedResponses) {
...
+          if (filterItem.item.description === 'region') {
```
burgwyn: Let's add Skill Code too

src/actions/filters.js
```diff
@@ -38,31 +38,49 @@ export function filtersFetchData(items, queryParams, savedResponses) {
...
+
+      // set any custom descriptions
+      responses.filters.forEach((filterItem, i) => {
+        filterItem.data.forEach((filterItemObject, j) => {
+          if (filterItem.item.description === 'region') {
+            responses.filters[i].data[j].custom_description =
+              `${filterItemObject.long_description}
+              (${filterItemObject.short_description})`;
+          }
+          if (filterItem.item.description === 'skill') {
```
burgwyn: else if?

PR #730, opened 2017-10-05
Favorites - Updating labels and adding second snapshot

### LINKED ISSUE #725: Favorites title attribute
At some point during the merges among `dev`, `4-month-demo`, and `sprint-8`, some thing must have gotten squashed in the `Favorite` component. The `title` attribute text should be updated so that it accurately represents what action is taking place.

Currently: "Remove from Favorites" and "Favorite from Favorites"
Should be: "Remove from Favorites" and "Add to Favorites"

burgwyn:
This closes #725

### DIFF
--- a/src/Components/Favorite/Favorite.jsx
+++ b/src/Components/Favorite/Favorite.jsx
@@ -24,7 +24,8 @@ class Favorite extends Component {
   }
 
   render() {
-    const text = this.getSavedState() ? 'Remove from Favorites' : 'Add to Favorites';
+    const text = this.getSavedState() ? 'Remove' : 'Favorite';
+    const title = this.getSavedState() ? 'Remove from Favorites' : 'Add to Favorites';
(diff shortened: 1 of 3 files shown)


src/Components/Favorite/Favorite.jsx
```diff
@@ -35,7 +35,7 @@ class Favorite extends Component {
       // eslint-disable-next-line jsx-a11y/no-static-element-interactions
       <div
         tabIndex="0"
-        title={`${text} from favorites`}
+        title={text}
```
mjoyce91: We can certainly change both the visible text as well as the `title` attribute to be the same thing, but originally we simply showed "Favorite" and "Remove" as the visible text, and then used "Remove/Add from/to Favorites" as the `title` attribute. The title attribute comes into play with the condensed cards, because we hide the text next to the heart icon, but need a `title` for accessibility.

burgwyn:
Good catch, fixed

Issue #634, opened 2017-09-21
UI - Submit Bid List


burgwyn:
As a bidder
I need a means to submit my bid list
So that I can confirm interest in a position

## Acceptance Criteria
- A mechanism for adding a single position to the bid list
- A mechanism for removing a single position from the bid list
- A view that displays all current positions on the bid list
- A means for a bidder to submit their final bid to TalentMAP for review
- A confirmation of submission
- Ensure a minimum of 1 bid and parameter maximum of bids are submitted

mjoyce91:
@afrasier - I'm guessing the API will check that the Bid List meets the parameter minimum/maximum upon attempted submission, correct?

afrasier:
Yeah, I'm thinking we'll hardcode that number for now and when we get a story to allow for changing it I'll factor it out into something the front end could alter.

Basically, you'll call something like PUT `/api/v1/bidlist/position/{id}/` to add a position to a bidlist, then you'll call `/api/v1/bidlist/submit/` to submit the current contents of the bidlist. You can add extra bids to the list after the submission, but they won't be marked as submitted unless you submit the list again. (Submitting will be an idempotent process, so it won't affect previously submitted bids)

PR #732, opened 2017-10-05
Add to Bid List

mjoyce91:
Addresses #640, #641, and #644, which are subtasks of #634. Adds redux actions to retrieve the user's bid list and toggle a position from their bid list. Adds a button to the Position Details to integrate these actions.

### DIFF
--- a/src/Components/BidListButton/BidListButton.test.jsx
+++ b/src/Components/BidListButton/BidListButton.test.jsx
@@ -5,7 +5,9 @@ import toJSON from 'enzyme-to-json';
 import BidListButton from './BidListButton';
 
 describe('BidListButtonComponent', () => {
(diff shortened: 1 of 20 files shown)


src/Components/BidListButton/BidListButton.test.jsx
```diff
@@ -1,19 +1,69 @@
...
+        toggleBidPosition={spy}
+        compareArray={[{ id: 2 }]}
+        isLoading={false}
+      />,
+    );
+    wrapper.find('button').simulate('click');
+    sinon.assert.calledOnce(spy);
+  });
+
   it('matches snapshot', () => {
```
burgwyn: Let's have two snapshot to capture the on and off states

PR #734, opened 2017-10-06
Display User's Bid List 

mjoyce91:
Addresses #642 and #643 by adding a section to the Profile that displays all positions the user has added to their bid list. Each results card includes a link to the position and a button to delete the position from their bid list.

PR #735, opened 2017-10-06
Bid list submission action

mjoyce91:
Relies on #734.

Keeping PRs in smaller chunks. This PR adds redux action for submitting a bid by position ID.

### DIFF
--- a/src/Components/ProfilePage/ProfilePage.jsx
+++ b/src/Components/ProfilePage/ProfilePage.jsx
@@ -1,10 +1,11 @@
 import React from 'react';
-import { Route } from 'react-router-dom';
+import { Route, Switch } from 'react-router-dom';
 import PropTypes from 'prop-types';
 import ProfileNavigation from '../ProfileNavigation';
 import FavoritePositions from '../FavoritePositions';
 import SavedSearches from '../SavedSearches';
 import ProfileLanding from '../ProfileLanding';
+import BidList from '../BidList';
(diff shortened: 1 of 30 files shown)


src/Components/ProfilePage/__snapshots__/ProfilePage.test.jsx.snap
```diff
@@ -5,7 +5,7 @@ exports[`ProfilePageComponent matches snapshot 1`] = `
   className="usa-grid-full"
 >
   <h1>
-    Hello, john!
+    Hello, ...!
```
burgwyn: We need to update this

src/Components/ProfilePage/ProfilePage.jsx
```diff
@@ -24,43 +27,61 @@ cloneSavedSearchSuccess }) => (
...
-              deleteSavedSearchSuccess={deleteSavedSearchSuccess}
-              cloneSavedSearch={cloneSavedSearch}
-              cloneSavedSearchIsLoading={cloneSavedSearchIsLoading}
-              cloneSavedSearchHasErrored={cloneSavedSearchHasErrored}
-              cloneSavedSearchSuccess={cloneSavedSearchSuccess}
-            />
-          )
-        }
-      />
+      <Switch>
```
burgwyn: Can we test these profile routes from outside the profile page?  Are the accessible from, say, the home page?

Issue #772, opened 2017-10-17
BidListCard - Refactor SAVED_SEARCH_PARENT_OBJECT


burgwyn:
Create a bid-specific shape

Issue #768, opened 2017-10-17
Position Details - Fix Bid Button Alignment


burgwyn:
Design aligns the button as so:
![design](https://user-images.githubusercontent.com/661204/31642842-3c8281c2-b2bb-11e7-818f-8d65747115b0.png)

Actual alignment:
![execution](https://user-images.githubusercontent.com/697848/31627421-406a0a2a-b27b-11e7-8f68-dc4acaf46a2f.png)

mjoyce91:
This one is actually dependent on the length of the Position's description. Examples:

<img width="967" alt="screen shot 2017-10-19 at 12 24 12 pm" src="https://user-images.githubusercontent.com/11576347/31782263-77c0e7ba-b4c8-11e7-97b1-ea14db27fea7.png">

<img width="927" alt="screen shot 2017-10-19 at 12 24 18 pm" src="https://user-images.githubusercontent.com/11576347/31782264-77e47176-b4c8-11e7-82ee-01b16c9465e5.png">

krisdelacruz:
@mjoyce91 I have a solution for the long descriptions. We can discuss.

Issue #767, opened 2017-10-17
BidListResultsList - Clean up imports


burgwyn:
No need for '*', `import { BID_LIST } ...`

Issue #766, opened 2017-10-17
Logo - Focus is off


burgwyn:
Explore styles and resolve focus 

![image](https://user-images.githubusercontent.com/661204/31642718-627f7cd2-b2ba-11e7-8b9f-adf15e9b7610.png)

PR #776, opened 2017-10-17
Sprint 9 Fixes

burgwyn:
This addresses:
#763
#764 
#765
#766 
#767
#768
#769 
#772

### DIFF
--- a/src/Components/Header/Header.jsx
+++ b/src/Components/Header/Header.jsx
@@ -55,11 +55,9 @@ export class Header extends Component {
         <div className="usa-navbar">
           <button className="usa-menu-btn">Menu</button>
           <div className="usa-logo" id="logo">
-            <div className="usa-logo-text">
-              <Link to="/">
-                <img src={logo} width="200" alt="TalentMAP logo" />
-              </Link>
-            </div>
+            <Link to="/">
+              <img src={logo} alt="TalentMAP logo" />
+            </Link>
           </div>
         </div>
         <nav className="usa-nav">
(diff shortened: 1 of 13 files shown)


src/Components/Header/Header.jsx
```diff
@@ -55,11 +55,9 @@ export class Header extends Component {
         <div className="usa-navbar">
           <button className="usa-menu-btn">Menu</button>
           <div className="usa-logo" id="logo">
-            <div className="usa-logo-text">
```
mjoyce91: Removing this class breaks some styles in responsive views.

With class:
screen shot 2017-10-19 at 12 20 31 pm

Without class:
screen shot 2017-10-19 at 12 20 25 pm

src/login/index.jsx
```diff
@@ -22,8 +22,8 @@ export class Login extends Component {
 
   // Remember, Redux Form passes the form values to our handler
   // In this case it will be an object with `username` and `password`
-  submit(values) {
-    this.props.loginRequest(values);
+  submit({ username, password }) {
```
mjoyce91: 👍

src/login/index.test.jsx
```diff
@@ -29,18 +29,6 @@ describe('Login', () => {
     expect(wrapper).toBeDefined();
   });
 
-  it('can call the submit function', () => {
```
mjoyce91: Why'd we remove this test?

src/sass/_details.scss
```diff
@@ -128,5 +128,5 @@
 .offset-bid-button-container {
   left: 63%;
   position: absolute;
-  top: 79%;
+  top: 85%;
```
mjoyce91: This works, but there's probably a way to get the button consistently on the middle of the edge regardless of the parent container's height (see my comment on  #768).

src/login/index.test.jsx
burgwyn: The test right below is similar and more substantive

mjoyce91: Ah didn't see that. Yeah that makes sense to test the actual interactions.

PR #809, opened 2017-10-25
Node Security Fix

burgwyn:
Node has issued a security patch, 6.11.5, to fix a vulnerability in the underlying `zlib` library.  More info [here](https://nodejs.org/en/blog/release/v6.11.5/)

Fix updates:
- Base Docker image
- Documentation

Excludes:
- CircleCI docker compose (image isn't ready yet)


Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM node:6.11
+FROM node:6.11.5
```
jseppi: The previously-specified `FROM node:6.11` should pick up patch releases already, right? I think it would just require a full rebuild without cache: `docker-compose build --no-cache`

jseppi:
This got me to thinking that we might want some simple "meta" tests to make sure runtime versions are all in-sync since they're specified in multiple places. Basically I'm thinking of a test that looks in the various Dockerfiles, docker-compose.ymls, README, package.json, etc. to make sure the desired node (and maybe postgres) versions are present across all files. We did something like in CALC: https://github.com/18F/calc/blob/develop/hourglass/tests/test_configuration.py

Also, rather that specifying the `nvm use 6.11.5` in the README, perhaps committing an `.nvmrc` file to the repo would be easier.

Anyways, this change is fine, so these comments are more just food for thought.

DEPLOY.md
```diff
@@ -15,12 +15,12 @@ sudo yum install git -y
...
 ## Install nvm
 
-Install Node version 6.11.0
+Install Node version 6.11.5
 
 ```
 curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.32.0/install.sh | bash
 . ~/.nvm/nvm.sh
-nvm install 6.11.0
+nvm install 6.11.5
```
jseppi: Since we now have `.nvmrc` file, I think this can be simplified to just `nvm install && nvm use` since it'll automatically pick up the version in `.nvmrc` :)

And you can probably remove the version callout up on line 18, too.

burgwyn: Will do

DEPLOY.md
```diff
@@ -15,12 +15,13 @@ sudo yum install git -y
 
 ## Install nvm
 
-Install Node version 6.11.0
+Install Node version 6.11.5
```
jseppi: Maybe change to "Install Node version specified in .nvmrc"

Issue #789, opened 2017-10-18
Position Listing for Post


burgwyn:
As a bidder
I want to know all the available positions for a post
So that I can search for a position based on post

## Acceptance Criteria
* A view listing all positions that fall under a post
* Data includes:
  * Position Number
  * Position Title
  * Post
  * Bureau
  * Skill code
  * Grade

Issue #788, opened 2017-10-18
Position Listing for a Mission


burgwyn:
As a bidder
I want to know all the available positions for a mission
So that I can search for a position based on mission

## Acceptance Criteria
* A view listing all positions that fall under a mission (country)
* Data includes:
  * Position Number
  * Position Title
  * Post
  * Bureau
  * Skill code
  * Grade

PR #810, opened 2017-10-26
Post and Mission Autocomplete Filters

mjoyce91:
Addresses #788 and #789 by integrating the `react-autosuggest` package for use with two new filters on the search results page: Post and Mission. User can filter search results by Post or Mission using a subset of the Post and Mission data as multi-select filters, or by searching for them using an autocomplete text input. Autocomplete results are queried from the API asynchronously.

Relies on https://github.com/18F/State-TalentMAP-API/pull/190 if it hasn't already been merged into `dev`.


src/sass/_variables.scss
```diff
@@ -19,4 +21,5 @@ $avatar-dropdown-account-segment-color: #e1e1e1;
 
 // z-index values
 // we'll use this section to keep track of and order them
+$autosuggest-suggestions-container-open-z: 2;
```
burgwyn: 👍

src/utilities.js
```diff
@@ -182,3 +182,8 @@ export const ifEnter = (e) => {
 
 // convert a query object to a query string
 export const formQueryString = queryObject => queryString.stringify(queryObject);
+
+// remove duplicates by property
+export const removeDuplicates = (myArr, prop) => (
```
burgwyn: This needs a test.  It gets covered elsewhere, but we should still test it

src/Components/AutoSuggest/RenderSuggestion/RenderSuggestion.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const RenderSuggestion = ({ suggestion }) => (
+  <div className="render-suggestion">
```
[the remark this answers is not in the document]
mjoyce91: We don't have to name it `RenderSuggestion` - that's just the prop name `react-autosuggest` uses for the template so I matched that. `SuggestionContainer` sounds good.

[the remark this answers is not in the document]
burgwyn: I'm having second thoughts about the name `SuggestionContainer` since containers are a thing elsewhere.  See if you can some up with something better.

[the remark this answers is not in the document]
mjoyce91: Right, we don't really control the overarching container of suggestions. `SuggestionChoice` works.

src/Components/AutoSuggest/SuggestionChoice/SuggestionChoice.jsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
```
burgwyn: 👍

src/Constants/EndpointParams.js
```diff
@@ -22,7 +24,14 @@ export const VALID_PARAMS = [
...
+  ENDPOINT_PARAMS.post,
   'q',
 ];
 
+export const ASYNC_PARAMS = [
+  ENDPOINT_PARAMS.mission,
+  ENDPOINT_PARAMS.post,
+];
+
 export default ENDPOINT_PARAMS;
```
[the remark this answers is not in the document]
mjoyce91: Removed.

src/utilities.js
mjoyce91: Added.

src/Components/ResultsPage/ResultsPage.jsx
```diff
@@ -1,8 +1,6 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import { POSITION_SEARCH_RESULTS, EMPTY_FUNCTION, USER_PROFILE, SAVED_SEARCH_MESSAGE,
-  SORT_BY_PARENT_OBJECT, PILL_ITEM_ARRAY, ACCORDION_SELECTION_OBJECT, FILTER_ITEMS_ARRAY,
-  SAVED_SEARCH_OBJECT } from '../../Constants/PropTypes';
+import * as PROP_TYPES from '../../Constants/PropTypes';
```
burgwyn: Let's avoid `*`

mjoyce91: I've been moving that way once the import list gets larger. Do you suggest doing that so that it's more clear which elements the component is actually using?

burgwyn: I know it is a pain, but I think we should

mjoyce91: No problem, will update.

mjoyce91: I opted to only do this for Components and Containers. For example, when testing an action or reducer, I think it's implied that all functions from the source will be used. What do you think?

Issue #814, opened 2017-10-27
Incorporate sass-lint


burgwyn:
The `scss-lint` team is [actively discouraging adoption](https://github.com/brigade/scss-lint).  I had another project team kick off over the summer and we opted for `sass-lint`.  

## Pros
* This looks like a straightforward switch

## Cons
* We lose CodeClimate integration
  * Need to run `sass-lint` as part of the CircleCI build to compensate

PR #815, opened 2017-10-27
Sprint 10 CodeClimate Fixes

burgwyn:
There's some questionable stuff in here that can be easily explained...

## ESLint
* Spacing fix
* Disabling `jsx-a11y` rule - I believe we're in compliance because we specify a `role`.  Additionally, we're way ahead of UI/UX on this.

## scss-lint
* So, I may have used `sass-lint` and not `scss-lint` locally
  * Recommending we move to `sass-lint` #814
* `_editor.scss` carries many dependencies from an external library

### DIFF
--- a/src/Components/TextEditor/TextEditor.jsx
+++ b/src/Components/TextEditor/TextEditor.jsx
@@ -41,7 +41,9 @@ export default class TextEditor extends Component {
     const { readOnly } = this.props;
     return (
       <div>
-        { /* eslint "jsx-a11y/no-static-element-interactions": 0 */ }
+        { /* outer div here exists to override styling & behavior of the interior component */ }
+        { /* this eslint rule seems to fire even though the role is defined */}
+        { /* eslint-disable-next-line jsx-a11y/no-static-element-interactions */ }
         <div role="textbox" tabIndex="0" className={readOnly ? '' : 'editor'} onClick={this.focus}>
           <Editor
             editorState={this.state.editorState}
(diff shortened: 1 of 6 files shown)


src/Components/TextEditor/TextEditor.jsx
```diff
@@ -41,6 +41,7 @@ export default class TextEditor extends Component {
     const { readOnly } = this.props;
     return (
       <div>
+        { /* eslint "jsx-a11y/no-static-element-interactions": 0 */ }
```
jseppi: does `eslint-disable-next-line jsx-a11y/no-static-element-interactions` work here instead?

src/sass/_editor.scss
```diff
@@ -1,12 +1,13 @@
+// scss-lint:disable all
```
jseppi: maybe a comment for why it is disabled would be nice (ie because it contains mostly overrides for a module's built-in styling)

src/Components/TextEditor/TextEditor.jsx
```diff
@@ -41,6 +41,9 @@ export default class TextEditor extends Component {
     const { readOnly } = this.props;
     return (
       <div>
+        { /* outer div here exists to override styling & behavior of the interior component */ }
+        { /* this eslint rule seems to fire even though the role is defined */}
```
mjoyce91: This is a known issue CodeClimate. See the `Favorite`, `CompareCheck` and `ResetFilter` components:

```At the time of writing, CodeClimate's version of eslint-a11y-plugin
did not take role="button" into account with the following error```

src/sass/_editor.scss
mjoyce91: I think we should also use `scss-lint:disable SelectorFormat`. We're actually ignoring the color literals when we `disable all` (but that's totally on me for forgetting to externalize those to `_variables`).

Issue #858, opened 2017-10-31
Move the debounced function inside the AutoSuggest class


mjoyce91:
Can move the debounced function into the class:

```js
constructor(props) {
  // ...
  this.debounced = debounce(() => {});
}
```

PR #869, opened 2017-10-31
Move debounced function into class

mjoyce91:
#858


src/Components/AutoSuggest/AutoSuggest.jsx
```diff
@@ -27,6 +23,10 @@ export default class AutoSuggest extends Component {
       value: '',
       suggestions: [],
     };
+
+    // Configure our debounce function.
+    // It needs to be global so we can cancel any pending requests while the user types.
```
jseppi: comment is out of sync with reality now. could say something like `// create an instance attribute for storing a reference to debounced requests`

Issue #864, opened 2017-10-31
Add tests for TextEditor component


mjoyce91:
Test that `editorState` and `editorStateCopy` are set to expected values.

PR #873, opened 2017-10-31
Add TextEditor test to check for initial value equality

mjoyce91:
#864

### DIFF
--- a/src/Components/TextEditor/TextEditor.test.jsx
+++ b/src/Components/TextEditor/TextEditor.test.jsx
@@ -72,4 +72,46 @@ describe('TextEditorComponent', () => {
       .getCurrentContent().getPlainText();
     expect(instanceEditorState).toBe(instanceEditorStateCopy);
   });
+
+  it('can properly maintain the value of the editorState and editorStateCopy after canceling', () => {
+    const wrapper = shallow(
+      <TextEditor
+        id={id}
+        onSubmitText={() => {}}
+        initialText="test"
+      />,
+    );
+    // regular and copied state should initialize as different objects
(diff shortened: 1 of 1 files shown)


jseppi:
Can you test that they are the expected values after clicking submit and cancel? Basically what I was trying to get at was testing that clicking submit or cancel results in passing the correct `editorState` prop down to the `Editor` component.

mjoyce91:
@jseppi - ah I see what you mean. Mocking the way that draftjs forms immutable objects might be tricky, but I'll add a test to ensure that `editorState` and `editorStateCopy` match after invoking `submit` or `cancel`.

jseppi:
hrm, could you grab the `Editor` and check its props?

Something like 

```js
// after cancel
expect(wrapper.find('Editor').props('editorState')).toBe(originalEditorState);

// after submit
expect(wrapper.find('Editor').props('editorState')).toBe(newEditorState);
```

mjoyce91:
beat me to it

mjoyce91:
I check for inequality initially just as a sanity check

burgwyn:
Guys... 🎃

jseppi:
We don’t want to be haunted by the spectre of spoooooky tests!

mjoyce91:
Weird, if I run a wrapper.find('Editor'), I get `Method “props” is only meant to be run on a single node. 0 found instead.`, but if I throw an `id="a"` on `<Editor>`, I can find by `"#a"`.

jseppi:
That's odd o_O
You can do `console.log(wrapper.debug())` to see exactly what the wrapper looks like.

mjoyce91:
Weird, its name changes to `PluginEditor`.

Issue #860, opened 2017-10-31
Handle invalid property names in AutoSuggest templates


mjoyce91:
When retrieved objects properties contain `null` instead of an expected value, unexpected behavior can occur from the `AutoSuggest` dropdown. Handle the possibility of non-existstent properties.

PR #890, opened 2017-11-01
Filter null locations in post autocomplete action

mjoyce91:
In some cases, posts with `"location": null` break the autosuggest template. While results such as this should not be returned in production, it's good to handle for the possibility so that the autosuggest list doesn't break when it does happen.

#860

### DIFF
--- a/src/actions/autocomplete/postAutocomplete.js
+++ b/src/actions/autocomplete/postAutocomplete.js
@@ -36,7 +36,12 @@ export function postSearchFetchData(query) {
     )
       .then((response) => {
         dispatch(postSearchIsLoading(false));
-        return response.data.results;
+        let filteredResults = [];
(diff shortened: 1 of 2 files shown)


src/actions/autocomplete/postAutocomplete.js
```diff
@@ -36,7 +36,12 @@ export function postSearchFetchData(query) {
     )
       .then((response) => {
         dispatch(postSearchIsLoading(false));
-        return response.data.results;
+        let filteredResults = [];
+        if (response.data && response.data.results) {
```
jseppi: I think axios guarantees that `response.data` is defined, so no need to check it: https://www.npmjs.com/package/axios#response-schema

you could even change the function signature to

```jsx
.then(({ data }) => {
  // ...
```

PR #894, opened 2017-11-02
Profile page menu

mjoyce91:
Adds a side-nav/menu to the Profile page based on wireframes. Keeps profile page otherwise untouched for the most part, with the exception of adding a `max-height` and `overflow-y` to work well with the new menu.

### DIFF
--- a/src/Components/ProfileMenu/ProfileMenu.jsx
+++ b/src/Components/ProfileMenu/ProfileMenu.jsx
@@ -1,0 +1,31 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import NavLinksContainer from './NavLinksContainer';
+import NavLink from './NavLink';
+
+const ProfileMenu = ({ currentPath }) => (
+  <div className="usa-grid-full profile-menu">
+    <div className="menu-title">
+      Menu
+    </div>
+    <NavLinksContainer>
+      <NavLink title="Home" iconName="home" link="/profile/" isHighlighted={currentPath === '/profile/'} />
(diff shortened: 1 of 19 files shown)


src/Components/ProfileMenu/ProfileMenu.jsx
```diff
@@ -0,0 +1,31 @@
...
+import NavLinksContainer from './NavLinksContainer';
+import NavLink from './NavLink';
+
+const ProfileMenu = ({ currentPath }) => (
+  <div className="usa-grid-full profile-menu">
+    <div className="menu-title">
+      Menu
+    </div>
+    <NavLinksContainer>
+      <NavLink title="Home" iconName="home" link="/profile/" isHighlighted={currentPath === '/profile/'} />
```
burgwyn: `currentPath === '/profile/'` strikes me as something that could be bug-prone

mjoyce91: Yeah, something I just noticed is if you navigate directly to `/profile/searches` (no trailing slash), then it doesn't highlight. `Link` has a built in prop to do this, but I need to perform it to the parent container.

burgwyn: We may want to keep a `Map` in the parent to ensure unique values for the links

mjoyce91: https://github.com/ReactTraining/react-router/blob/master/packages/react-router/modules/matchPath.js

PR #895, opened 2017-11-03
Sprint 11 - CI

burgwyn:
Changing our CircleCI process to use a custom pa11y script that includes authentication in lieu of anonymous `pa11y-ci`.

This includes a lot of work from @jesse-russell and @mjoyce91


mjoyce91:
Something I realized for this is that we're going to need to pass a real API token to properly test everything. It used to be the case that any old token like `"faketoken123"` would give you access to the site. But, some of the routes we're now hitting require authentication. For example, loading a position details page requires auth since we need things like `is_editable_by_user`, etc.

jseppi:
Is this PR still relevant and in need of review?

burgwyn:
Closing as no longer applicable

PR #896, opened 2017-11-03
Profile dashboard grid

mjoyce91:
Adds the base container and grid for the profile dashboard. Uses dummy components which will eventually be filled with data. Used with a `connect`ed container, but not pulling any thing from redux state (will be one of the next steps).

Dependent on #894

### DIFF
--- a/src/Components/ProfileMenu/NavLink/NavLink.test.jsx
+++ b/src/Components/ProfileMenu/NavLink/NavLink.test.jsx
@@ -47,7 +47,7 @@ describe('NavLinkComponent', () => {
     expect(wrapper).toBeDefined();
   });
 
-  it('can apply link-highlighted class when isHighlighted is true', () => {
+  it('applies link-highlighted class when isHighlighted is true', () => {
     const wrapper = shallow(
       <NavLink
         title="test"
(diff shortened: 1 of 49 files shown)


src/Components/ProfileMenu/NavLink/NavLink.test.jsx
```diff
@@ -0,0 +1,123 @@
...
+        <NavLink
+          title="test3"
+          link="/profile/searches/"
+        />
+      </NavLink>,
+    );
+    expect(wrapper).toBeDefined();
+  });
+
+  it('can apply link-highlighted class when isHighlighted is true', () => {
```
burgwyn: Let's remove "can" from these descriptions.  "It applies..." better communicates what we're testing.

src/Components/ProfilePage/ProfilePage.jsx
```diff
@@ -1,25 +1,26 @@
...
+    <ProfileMenu currentPath={currentPath} />
+    <div className="usa-grid-full profile-content-container">
+      <div className="hello-greeting">
+        {
+          `Hello, ${user.user.username}`
+        }
+      </div>
       <Switch>
         <Route path="/profile" exact component={ProfileLanding} />
+        <Route path="/profile/dashboard" exact component={Dashboard} />
```
burgwyn: What is the reasoning for `exact` here but not elsewhere?  Is it child routes?

mjoyce91: May have just copy and pasted the element above since this was a new route. We'd only need it if `/dashboard` had child routes.

PR #898, opened 2017-11-06
Profile dashboard user and position containers

mjoyce91:
Covers #823 & #824 by adding two sections to the Profile Dashboard - the user's profile information, and their position information. Adds a new action to pull data from `/profile/assignments`.

Dependent on #894 and #896.


src/Components/ProfileDashboard/PositionInformation/PositionInformation.jsx
```diff
@@ -1,9 +1,37 @@
...
+        title="Position Title"
+        content={assignment.position || NO_ASSIGNMENT_POSITION}
+      />
+      <InformationDataPoint title="Skill Code" content="Medical Technology (6145)" />
+    </div>
   </div>
 );
 
+PositionInformation.propTypes = {
+  assignment: PropTypes.shape({}).isRequired,
```
burgwyn: Do we have a shape in-place?

src/Components/ProfileDashboard/__snapshots__/ProfileDashboard.test.jsx.snap
```diff
@@ -4,47 +4,11 @@ exports[`ProfileDashboardComponent matches snapshot when editor is shown 1`] = `
...
-    >
-      <BidList />
-    </div>
-    <div
-      className="usa-width-one-whole user-dashboard-section inbox-section"
-    >
-      <Inbox />
-    </div>
-  </div>
+  <Spinner
```
burgwyn: This needs to reflect the loaded state

src/Containers/Dashboard/Dashboard.jsx
```diff
@@ -11,31 +13,47 @@ class DashboardContainer extends Component {
   }
 
   componentWillMount() {
+    this.props.fetchAssignment();
   }
 
   render() {
+    // eslint-disable-next-line
```
burgwyn: Do we need this?

src/__mocks__/userObject.js
```diff
@@ -0,0 +1,10 @@
+const userObject = {
+  skill_code: 'INFORMATION MANAGEMENT (2880)',
+  user: {
+    first_name: 'John',
+    last_name: 'Doe',
+    email: 'jdoe@state.gov',
```
burgwyn: doej

src/Containers/Dashboard/Dashboard.jsx
```diff
@@ -11,31 +13,47 @@ class DashboardContainer extends Component {
...
+  userProfile: USER_PROFILE.isRequired,
+  userProfileIsLoading: PropTypes.bool.isRequired,
+  fetchAssignment: PropTypes.func.isRequired,
+  assignment: PropTypes.shape({}).isRequired,
+  assignmentIsLoading: PropTypes.bool.isRequired,
 };
 
 DashboardContainer.defaultProps = {
+  userProfile: DEFAULT_USER_PROFILE,
+  userProfileIsLoading: true,
```
burgwyn: We default to `false` in other Containers.  I think this may be breaking the snapshot.

src/__mocks__/userObject.js
mjoyce91: 🐶

src/Components/ProfileDashboard/ProfileDashboard.test.jsx
```diff
@@ -11,13 +11,13 @@ describe('ProfileDashboardComponent', () => {
...
-        isLoading
-        assignmentIsLoading
+        isLoading={false}
+        assignmentIsLoading={false}
       />);
     expect(wrapper).toBeDefined();
   });
 
-  it('matches snapshot when editor is shown', () => {
+  it('matches snapshot when loading', () => {
```
burgwyn: I don't think we need this kind of test since loading is fairly uniform across the app

mjoyce91: Gotcha... I can see this view having a lot `bidListIsLoading`, `assignmentsIsLoading`, `profileIsLoading`, etc. variable checks, which is why I wanted to capture a snapshot to ensure that the view loads when all `[x]isLoading` variables are true.

burgwyn: That makes sense and probably worth leaving it in for that reason

PR #902, opened 2017-11-08
Profile Dashboard Bid List

mjoyce91:
Addresses #825 by adding the Bid List section to the Profile Dashboard. Applies some business rules to determine styles (colors, icon) for the bid, based on its `status`. 

Aspects that are static:
- "All bids due by" date
- "More info" button

Other notes:
- There is no "added_date" for bids returned by the API, so we can't show the "Added on..." text found in the wireframes. Instead, we just show "Updated on..."

### DIFF
--- a/src/Components/BidListResultsCard/BidListResultsCard.jsx
+++ b/src/Components/BidListResultsCard/BidListResultsCard.jsx
@@ -1,7 +1,10 @@
 import React, { Component } from 'react';
-import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
-import { BID_OBJECT } from '../../Constants/PropTypes';
+import { BID_OBJECT, EMPTY_FUNCTION } from '../../Constants/PropTypes';
+import { SUBMITTED } from '../../Constants/BidStatuses';
+import { NO_POST } from '../../Constants/SystemMessages';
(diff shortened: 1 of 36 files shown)


src/Components/BidListResultsCard/BidListResultsCard.jsx
```diff
@@ -18,40 +21,32 @@ class BidListResultsCard extends Component {
...
-              <button
-                className="usa-button-secondary"
-                onClick={this.removeBidPosition}
-                title="Delete this position from your bid list"
-              >
-                Delete
-              </button>
-            </div>
+          // this section we'll only show if condensedView is false
+          !condensedView &&
```
burgwyn: These props are not required.  Should we construct this in a function with appropriate checks?

mjoyce91: I changed the functions to unrequired here since this component can be used with an optional `condensedView` prop, which hides any buttons (no buttons = no functions needed). Do you mean changing the check from `!condensedView` to `shouldShowBidActions()`?

burgwyn: My thought is there could be a scenario in which we show the condensed card but not the button functions.  We default `condensedView` to false and then `! condensedView`, so it seems possible.

Issue #758, opened 2017-10-11
Projected Vacancy View


burgwyn:
As a bidder
I want to know when a position is going to be vacant
So that I can plan to bid accordingly

## Acceptance Criteria
* Additional fields on position details
  * Projected vacancy date - estimated tour end date (TED)
  * Incumbent

PR #903, opened 2017-11-08
sprint-11 -> dev

mjoyce91:
Covers work from Sprint 11, including:

#758 - Projected Vacancy view (in position details page)
#820 - Profile Dashboard


src/Components/BorderedList/BorderedList.test.jsx
```diff
@@ -0,0 +1,41 @@
+import { shallow } from 'enzyme';
+import React from 'react';
+import toJSON from 'enzyme-to-json';
+import BorderedList from './BorderedList';
+
+describe('BidListResultsListComponent', () => {
```
burgwyn: 🍝 - BorderedList

src/Components/ProfileDashboard/Notifications/Notifications.jsx
```diff
@@ -0,0 +1,44 @@
+import React from 'react';
```
burgwyn: Let's rename to NotificationList.  Or call "Notification" something like "NotificationItem"

src/actions/autocomplete/postAutocomplete.js
```diff
@@ -1,7 +1,7 @@
-import axios from 'axios';
+import axios, { CancelToken } from 'axios';
 import api from '../../api';
 
-const CancelToken = axios.CancelToken;
+// const CancelToken = axios.CancelToken;
```
burgwyn: Did we mess up the merge here?

mjoyce91: We import switched to simply importing `{ CancelToken } from 'axios'`, so this must be an artifact of testing both ways out.

src/Components/ProfileDashboard/Notifications/Notifications.jsx
mjoyce91: I like NotificationItem, since it keep the dashboard sections named with the same convention.

PR #910, opened 2017-11-08
Change border color to green

mjoyce91:
The border for hand-shake-offered was the wrong color. This changes it from light blue to green.


mjoyce91:
Not sure what's going on with CircleCI. I can install all packages successfully with yarn locally.

PR #920, opened 2017-11-13
Fix copy-paste of test name in the ProfilePicture component

mjoyce91:


jseppi:
darn, I thought there was going to be some weird inverse/negative filter for profile pics

image

PR #921, opened 2017-11-13
'Hello username' -> 'Hello first_name last_name'

mjoyce91:


jseppi:
Hrm, something I guess to consider here is names that aren't ordered as `first last`, like Chinese names, which put the surname first. Do we know if DoS handle names like that in any specific way?

krisdelacruz:
@jseppi I can double check for the team. I believe this is also important on a CDO and AO level where they Search Bidders by the Last Name first. Great question, noted.

mjoyce91:
@burgwyn are we good to merge this?

PR #923, opened 2017-11-13
Add test for when titleOnBottom is true

mjoyce91:


src/Components/ProfileDashboard/InformationDataPoint/InformationDataPoint.test.jsx
```diff
@@ -28,4 +28,9 @@ describe('InformationDataPointComponent', () => {
     const wrapper = shallow(<InformationDataPoint content="content" />);
     expect(toJSON(wrapper)).toMatchSnapshot();
   });
+
+  it('matches snapshot when titleOnBottom is true', () => {
+    const wrapper = shallow(<InformationDataPoint content="content" titleOnBottom />);
```
jseppi: I think you should add a `title` to so you can see it rendered on the bottom of the snapshot

mjoyce91: Good point, doesn't do us much good without a title.

mjoyce91: Added title to both snapshot tests for good measure.

PR #931, opened 2017-11-14
Use min-height instead of overflow-y

mjoyce91:


jseppi:
hrm, even with this change there are still double scroll bars. Is there any way the entire page height could expand to fit the `profile-content-container`? Ie, so that the footer area is pushed down as needed, and the side nav area's height also expands all the way to the footer.

screen shot 2017-11-14 at 7 57 55 am

dual_scroll2

mjoyce91:
The Notifications container is scrollable, but this change should have made your `profile-content-container` un-scrollable, which it appears to still be doing in your screen cap 🤔. As for the footer, that should be fine - however, right now the footer's class contains top padding, so I'll need to move the padding outside of the footer and into the containers for the other pages.

src/sass/_containers.scss
```diff
@@ -0,0 +1,3 @@
+.content-container {
+  padding: 0 0 50px;
```
jseppi: Maybe just `padding-bottom: 50px;` so that any any potential overriding rules don't need to be as specific? Totally up to you -- CSS architecture is definitely not a strong point of mine.

mjoyce91:
@burgwyn - could we re-merge? Changed `padding` to `padding-bottom` based on feedback from @jseppi

PR #935, opened 2017-11-14
Abstract common mocks to a single file

### LINKED ISSUE #933: Abstract common code for actions tests
Create a file that abstracts common imports and logic used in actions.

```js
import configureMockStore from 'redux-mock-store';
import thunk from 'redux-thunk';
import axios from 'axios';
import MockAdapter from 'axios-mock-adapter';

export default function setupAsyncMocks() {
  const middlewares = [thunk];
  const mockStore = configureMockStore(middlewares);
  const mockAdapter = new MockAdapter(axios);

  return { mockStore, mockAdapter };
}
```

https://github.com/18F/State-TalentMAP/pull/915/files/505e3284ab5f7f203da6fdfb7924df6ae783e09f#diff-0c37385dcfd19e198c7302fd640bccdf

mjoyce91:
Closes #933.

### DIFF
--- a/src/actions/setupAsyncMocks.js
+++ b/src/actions/setupAsyncMocks.js
@@ -1,0 +1,14 @@
+// This file is used by tests, so we can ignore no-extraneous-dependencies
+/* eslint-disable import/no-extraneous-dependencies */
+import configureMockStore from 'redux-mock-store';
+import thunk from 'redux-thunk';
+import axios from 'axios';
+import MockAdapter from 'axios-mock-adapter';
+
+export default function setupAsyncMocks() {
+  const middlewares = [thunk];
+  const mockStore = configureMockStore(middlewares);
+  const mockAdapter = new MockAdapter(axios);
+
+  return { mockStore, mockAdapter };
+}
--- a/src/actions/assignment.test.js
+++ b/src/actions/assignment.test.js
@@ -1,17 +1,11 @@
-import configureMockStore from 'redux-mock-store';
-import thunk from 'redux-thunk';
-import axios from 'axios';
-import MockAdapter from 'axios-mock-adapter';
+import setupAsyncMocks from './setupAsyncMocks';
 import * as actions from './assignment';
 import assignmentObject from '../__mocks__/assignmentObject';
 
-const middlewares = [thunk];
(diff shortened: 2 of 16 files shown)


src/actions/setupAsyncMocks.js
```diff
@@ -0,0 +1,14 @@
+// This file is used by tests, so we can ignore no-extraneous-dependencies
+/* eslint-disable import/no-extraneous-dependencies */
```
jseppi: 👍 

If we end up with more of these test-support files, we might want to have a separate directory (like `src/testSupport/`) or naming convention (like `xyz.test-support.js`) for them, and then add that pattern to `.eslintrc`'s rules config https://github.com/18F/State-TalentMAP/blob/b72e1b1f3454bbd55e637bd15fb46219457e1271/.eslintrc#L30

burgwyn: 👍

mjoyce91:
I'm going to come back to this one to bring code coverage back up.

jseppi:
I am guessing the coverage percent went down because overall lines of code was reduced quite a bit, not because there is less coverage overall.

burgwyn:
Agreed.  I ran things locally and the overall coverage picture loos good.  Lots of green.

mjoyce91:
It looks to actually be because we're no longer catching errors when ever we have two mocks that reference the same URL. The mockAdapter has to be reinstantiated, otherwise the original mock is used (ex: if one is a `200` and the next is a `404`, only the `200` is used).

jseppi:
ah, like https://github.com/18F/State-TalentMAP/pull/935/files#diff-9d0d8c16acc7606d12aaa4fc7796fa85L56 for example?

Is there like a `mockAdapter.clear()` or something that could be used?

mjoyce91:
Ah good call, I was just going to reinstantiate with:

```js
beforeEach(() => {
    mockAdapter = setupAsyncMocks().mockAdapter;
  });
```

But there is actually a `reset`, so we can instead do:
```js
beforeEach(() => {
    mockAdapter.reset();
  });
```

mjoyce91:
And yeah, that is an example of one

PR #915, opened 2017-11-09
dev -> staging

mjoyce91:
Covers work from Sprint 11, including:

#758 - Projected Vacancy view (in position details page)
#820 - Profile Dashboard

PR #957, opened 2017-11-15
Sprint 11 Test Coverage

mjoyce91:
Increases test coverage. **Not** directly related to #915.

One of the larger areas here was to move some of the customized templating logic found in the `filters` action into its own `helpers` file.


src/Components/ProfileDashboard/InformationDataPoint/InformationDataPoint.test.jsx
```diff
@@ -9,6 +9,11 @@ describe('InformationDataPointComponent', () => {
     expect(wrapper).toBeDefined();
   });
 
+  it('it is defined when titleOnBottom is true', () => {
```
burgwyn: extra 'it'

PR #961, opened 2017-11-15
StaticDevContent wrapper component

mjoyce91:
A component to wrap known static/hard-coded values in. Adds a toggle link to the header so that static content can be toggled on or off. When on, static content receives a red, inner border. When off, which is the default state, no visual change is applied.

I've opted to ignore this file from testing which is why I've left this PR in WIP. If we think it's worth adding tests, I will add some in. I chose not to because it will ultimately not be a part of the finalized and delivered application.

### DIFF
--- a/src/Components/StaticDevContent/StaticDevContent.jsx
+++ b/src/Components/StaticDevContent/StaticDevContent.jsx
@@ -1,3 +1,9 @@
+/*
+Use this component to wrap bits of static development
+content (like a hardcoded phone number) so that they
+can be easily detected during demos or user testing.
+*/
+
 import React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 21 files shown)


src/Components/StaticDevContent/StaticDevContent.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+
+const StaticDevContent = ({ children, showStaticContent }) => (
+  showStaticContent ?
+    <div style={{ boxShadow: '0px 0px 0px 2px red' }}>
```
jseppi: clever use of `boxShadow`!

Issue #991, opened 2017-11-20
Abstract query param update functions for use across containers


mjoyce91:

burgwyn:
Closing

PR #1013, opened 2017-11-21
CDO Bidder Portfolio

mjoyce91:
Initial work on the CDO Bidder Portfolio.

#976 - CDO Client Page
#977 - CDO Client Card View
#978 - Action to retrieve CDO's clients
#979 - CDO Client Card Listing
#987 - CDO client pagination
#991 - Abstract query param function

Additionally, this PR abstracts the queryParamUpdate function that's also used in the `Results` container. It also simplifies using the `PaginationWrapper` so that the total number of pages it should display is handled within the wrapper component, instead of having to calculate that number outside of the component.


src/Components/ProfileDashboard/InformationDataPoint/__snapshots__/InformationDataPoint.test.jsx.snap
```diff
@@ -2,7 +2,24 @@
...
+  </div>
+  <div
+    className="data-point-content"
+  >
+    content
+  </div>
+</div>
+`;
+
+exports[`InformationDataPointComponent matches snapshot when sideBySide is true 1`] = `
```
burgwyn: 👍

src/Components/ProfileDashboard/PositionInformation/__snapshots__/PositionInformation.test.jsx.snap
```diff
@@ -21,25 +21,29 @@ exports[`PositionInformationComponent matches snapshot 1`] = `
...
         />
       }
+      sideBySide={false}
       title="Start & End of Position"
       titleOnBottom={false}
     />
     <InformationDataPoint
       className=""
       content="Bureau of Western Hemispheric Affairs"
+      sideBySide={false}
```
burgwyn: Good to set these, but the default is false

src/Components/ProfileMenu/__snapshots__/ProfileMenu.test.jsx.snap
```diff
@@ -23,6 +23,76 @@ exports[`ProfileMenuComponent matches snapshot 1`] = `
         link="/profile/dashboard/"
         title="Dashboard"
       />
+      <withRouter(NavLink)
```
burgwyn: I think this should be hidden...

src/Components/ProfileMenu/__snapshots__/ProfileMenu.test.jsx.snap
```diff
@@ -23,6 +23,76 @@ exports[`ProfileMenuComponent matches snapshot 1`] = `
...
+    />
+    <withRouter(NavLink)
+      iconName="user"
+      title="Profile"
+    >
+      <withRouter(NavLink)
+        link="/profile/dashboard/"
+        title="Dashboard"
+      />
+      <withRouter(NavLink)
```
burgwyn: ... because we see it here

src/__mocks__/userObject.js
```diff
@@ -5,7 +5,7 @@ const userObject = {
     last_name: 'Doe',
     email: 'doej@state.gov',
   },
-  is_cdo: false,
+  is_cdo: true,
```
burgwyn: Let's make 2 objects in here: `bidderUserObject` and `cdoUserObject`

src/reducers/index.js
```diff
@@ -103,6 +104,9 @@ export default combineReducers({
   notificationsHasErrored,
   notificationsIsLoading,
   notifications,
+  bidderPortfolio,
```
burgwyn: We appear to be increasing the number of reducers

src/Components/ProfileMenu/__snapshots__/ProfileMenu.test.jsx.snap
mjoyce91: I should export the unwrapped connected

src/Components/ProfileDashboard/PositionInformation/__snapshots__/PositionInformation.test.jsx.snap
mjoyce91: I don't explicitly declare this in the `PositionInformation` component, so it must be rendering it in the snapshot since it's the default prop in the `InformationDataPoint` component. I actually didn't realize it would do this.

src/Components/ProfileMenu/__snapshots__/ProfileMenu.test.jsx.snap
mjoyce91: Oh, now that I've taken a look at this, I see that this occurs with the nested `NavLink` elements in this component. I'm not sure if there's any way to hide this, but we do need to use the connected `withRouter` component.

src/reducers/index.js
mjoyce91: Is there a better way to structure or organize them?

burgwyn: I've name spaced them on other projects.  I'll combine reducers into logical groups and then combine those combinations.

mjoyce91: That makes sense. Right now they are generally in groups of three (the response, whether it's loading, and whether it's errored), so they could be grouped like that. I'll create an issue.

Issue #989, opened 2017-11-20
UI - Create Filter Tabs for CDO Portfolio


mjoyce91:

burgwyn:
Closing

PR #1022, opened 2017-11-27
CDO Bidder Portfolio - Navigation, Search, Sort

mjoyce91:
Continued build out of the CDO Bidder Portfolio.

#988 - CDO Client Portfolio Search
#989 - CDO Client Portfolio Filter/Navigation Tabs
#990 - CDO Client Portfolio Sorting

Relies on #1013 

Waiting on real filters to be implemented on the API, which is why all navigation/filter tabs display the same results.

### DIFF
--- a/src/Components/BidderPortfolio/BidderPortfolioContainer/BidderPortfolioContainer.test.jsx
+++ b/src/Components/BidderPortfolio/BidderPortfolioContainer/BidderPortfolioContainer.test.jsx
@@ -28,9 +28,19 @@ describe('BidderPortfolioContainerComponent', () => {
     sinon.assert.calledOnce(spy);
   });
 
-  it('matches snapshot', () => {
+  it('matches snapshot when the all property is greater than zero', () => {
     const wrapper = shallow(<BidderPortfolioContainer
       bidderPortfolio={bidderListObject}
+      pageSize={8}
+      pageNumber={1}
+      queryParamUpdate={() => {}}
+    />);
+    expect(toJSON(wrapper)).toMatchSnapshot();
+  });
+
+  it('matches snapshot when the all property is zero', () => {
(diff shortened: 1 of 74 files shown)


src/Components/BidderPortfolio/BidderPortfolioContainer/BidderPortfolioContainer.test.jsx
```diff
@@ -28,7 +28,7 @@ describe('BidderPortfolioContainerComponent', () => {
     sinon.assert.calledOnce(spy);
   });
 
-  it('matches snapshot', () => {
+  it('matches snapshot when the all property is greater than zero', () => {
```
burgwyn: Can you clarify the description?

src/Components/BidderPortfolio/TopNav/Actions/ActionsButton/ActionsButton.jsx
```diff
@@ -0,0 +1,7 @@
+import React from 'react';
+
+const ActionsButton = () => (
+  <button className="white-button">Actions...</button>
```
burgwyn: Will this button be clickable?

mjoyce91: Yes, once we know what this button will actually do

src/Components/BidderPortfolio/BidderPortfolioContainer/BidderPortfolioContainer.test.jsx
mjoyce91: Looks like this was an artifact of when I was using the `all` property to determine results, but I've since changed it to simply look at `results.length`. I've updated the test name accordingly.

Issue #831, opened 2017-10-30
Primary Navigation Design Updates


krisdelacruz:

burgwyn:
Closing

PR #1042, opened 2017-11-30
Notifications and Inbox alerts

mjoyce91:
Addresses #831 by Inbox and Notifications alerts to the header. Because Inbox is not yet implemented, it does not update with any alerts. Notifications is a connected component and updates with the number of unread notifications upon page change.

### DIFF
--- a/src/actions/notifications.js
+++ b/src/actions/notifications.js
@@ -21,12 +21,51 @@ export function notificationsFetchDataSuccess(notifications) {
   };
 }
 
+export function notificationsCountHasErrored(bool) {
+  return {
+    type: 'NOTIFICATIONS_COUNT_HAS_ERRORED',
+    hasErrored: bool,
+  };
+}
+export function notificationsCountIsLoading(bool) {
+  return {
+    type: 'NOTIFICATIONS_COUNT_IS_LOADING',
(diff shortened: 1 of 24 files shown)


src/actions/notifications.js
```diff
@@ -21,12 +21,51 @@ export function notificationsFetchDataSuccess(notifications) {
...
+
+export function unsetNotificationsCount() {
+  return (dispatch) => {
+    dispatch(notificationsCountFetchDataSuccess(0));
+  };
+}
+
+export function notificationsCountFetchData() {
+  return (dispatch) => {
+    axios.get(`${api}/notification/?limit=1&is_read=false`, { headers: { Authorization: fetchUserToken() } })
```
burgwyn: Let's make the limit a parameter and default it to something like 5.  Do the same with `Is_read` - make a param and default to `false`

mjoyce91: `limit` here is 1 because we don't actually care about reading the notifications from this response. We just want to pull the `count`, which is the same regardless of `limit`, so we might as well reduce the response size/time.

burgwyn: I see your point.  I think we may want to revisit the approach later, perhaps with a service worker.

PR #1060, opened 2017-12-04
Add small checkbox class to be used in the bidder portfolio list

mjoyce91:
Adds a `small` prop the `Checkbox` component so that a smaller checkbox can be used for the Bidder Portfolio list view.


src/Components/CheckBox/CheckBox.test.jsx
```diff
@@ -37,6 +37,22 @@ describe('CheckBoxComponent', () => {
     expect(onCheckBoxClick.calledOnce).toBe(true);
   });
 
+  it('can apply the small class', () => {
```
burgwyn: Can we snapshot this too?  Also, can we check for the `tm-checkbox` on the default case?

src/Components/CheckBox/__snapshots__/CheckBox.test.jsx.snap
```diff
@@ -2,7 +2,7 @@
 
 exports[`CheckBoxComponent matches snapshot 1`] = `
 <div
-  className="usa-grid-full tm-checkbox"
+  className="usa-grid-full tm-checkbox "
```
burgwyn: Is this class already defined?

mjoyce91: What do you mean?

src/Components/CheckBox/CheckBox.test.jsx
mjoyce91: 👍

src/Components/CheckBox/__snapshots__/CheckBox.test.jsx.snap
burgwyn: `.tm-checkbox` is utilized elsewhere in the app but not defined in `_checkbox.scss`

mjoyce91: Ah good point. I'll look into it.

mjoyce91: So I haven't defined any global rules for `tm-checkbox` but I like keeping it as a class in case we want to apply any down the road.

mjoyce91:
Looks like circleci status is stuck

PR #1073, opened 2017-12-05
Use Spinner while Position Details page is loading

### LINKED ISSUE #1071: Show loading spinner when loading a position's details page
Currently just the text "Loading..." is displayed while a position details view is loaded. Would be nice to show the loading spinner on that page to be consistent with other views.

no-spinner-view-details

mjoyce91:
Closes #1071


src/Components/PositionDetails/PositionDetails.jsx
```diff
@@ -70,7 +70,7 @@ class PositionDetails extends Component {
             <Share identifier={details.id} />
           </div>
         </div>}
-        {isLoading && <Loading isLoading={isLoading} hasErrored={hasErrored} />}
```
jseppi: I think this is the only place the `<Loading>` component was still used. Can we remove it and associated tests and styling?

PR #1086, opened 2017-12-06
Change button class name

### LINKED ISSUE #1085: Use less descriptive class name for alternate button
The the class name `white-button` is a bit too specific since the color of this alternate may change in the future. Rename is to `tm-button-alt`.

mjoyce91:
Closes #1085


src/sass/_whiteButton.scss
```diff
@@ -1,18 +1,18 @@
-.white-button {
+.tm-button-alt {
```
jseppi: filename should be changed as well :)

mjoyce91: Moved to buttons.scss file to consolidate

src/sass/_buttons.scss
```diff
@@ -23,3 +23,22 @@
...
+  font-size: 16px;
+  font-weight: 400;
+
+  .tm-button-alt-icon {
+    color: $blue-primary;
+    margin-right: 5px;
+  }
+}
+
+.tm-button-alt:hover {
```
jseppi: you could nest this in `.tm-button-alt` instead:

```scss
.tm-button-alt {
   // ...
   &:hover {
     // ...
   }
}
```

Not a big deal though

PR #1089, opened 2017-12-06
PropTypes

mjoyce91:
Some fixes related to PropTypes, and moving a commonly used object to a variable within a component.


src/Components/PositionTitle/PositionTitle.jsx
```diff
@@ -108,7 +108,9 @@ class PositionTitle extends Component {
 
     const isAllowedToEdit = !!(details.description && details.description.is_editable_by_user);
 
-    const shouldShowBidStats = !!details.bid_statistics && !!details.bid_statistics[0];
+    const bidStatistics = details.bid_statistics[0];
```
jseppi: hrm, this will throw an error if `bid_statistics` is undefined or null. If it can never be undefined then you don't need the first part (`!!details.bid_statistics`) of the boolean expression just below this.

src/Constants/PropTypes.js
```diff
@@ -320,7 +318,14 @@ export const BIDDER_OBJECT = PropTypes.shape(
...
     skill_code: PropTypes.string,
     grade: PropTypes.string,
-    cdo: PropTypes.number,
+    cdo: {
+      username: PropTypes.string,
+      first_name: PropTypes.string,
+      last_name: PropTypes.string,
+      email: PropTypes.string,
+      phone_number: PropTypes.string,
+      is_cdo: PropTypes.bool,
```
jseppi: just want to verify that `is_cdo` should also be in here

mjoyce91: Yes, that's returned in the API response (thought it seemed redundant though).

src/Components/PositionTitle/PositionTitle.jsx
mjoyce91: I feel like a safe assumption would be that `bid_statistics` is always defined, but in rare cases it might be `null` (this some times occurs within my local dev instance). So what about:

```js
const bidStatistics = details.bid_statistics && Array.isArray(details.bid_statistics) ?
      details.bid_statistics[0] : null;
```

This way, we don't evaluate `details.bid_statistics[0]` and risk an error being thrown if `bid_statistics` is undefined or null.

jseppi: cool, that sounds reasonable. Given that `bid_statistics` will always be defined (but could be null), then I think you could remove that first part of the ternary condition:

```js
const bidStatistics = Array.isArray(details.bid_statistics) ? details.bid_statistics[0] : null;
```

mjoyce91: Oh good point 👍

src/Constants/PropTypes.js
```diff
@@ -320,7 +318,14 @@ export const BIDDER_OBJECT = PropTypes.shape(
     id: PropTypes.number,
     skill_code: PropTypes.string,
     grade: PropTypes.string,
-    cdo: PropTypes.number,
+    cdo: {
```
mjoyce91: Oops, this needs to be a `shape`, not a simple object.

jseppi: good self-catch ⚾️

PR #1102, opened 2017-12-08
Use removeElement prop on account dropdown

### LINKED ISSUE #1096: Account Dropdown does not fully render in Microsoft Edge
Using Microsoft Edge, the Account Dropdown does not render on click. It only renders on pressing the enter key. However, the correct class is applied in either case, and once another spot on the screen is clicked, the Account Dropdown appears briefly.

Windows 10
Edge 41.16299.15.0

dec-07-2017 13-03-52

mjoyce91:
Closes #1096 by using the `removeElement` prop to create and destroy the account dropdown on click, instead of simply applying a class with `display: none` to hide it.


mjoyce91:
Tests working on this one now - just a codeclimate issue for 29 > 25 lines.

PR #1112, opened 2017-12-12
Header Collapse

mjoyce91:
Restructures the header and search bar so that the search bar can be collapsed via the Search link in the header. Will need to review more specialized logic for when the search bar when and will appear based on when and where the user chose to display it. Currently, it displays as follows:

1. Displays by default on the home page
2. Forces display on the search results page
3. Hidden by default on all other pages


src/Components/Header/Header.jsx
```diff
@@ -2,123 +2,128 @@ import React, { Component } from 'react';
...
         requesting,
       },
     } = this.props;
 
-    let showLogin = (<Link to="login" id="login-desktop">Login</Link>);
+    let showLogin = true; // (<Link to="login" id="login-desktop">Login</Link>);
+    let showAccountDropdown = false;
     let signedInAs = null;
-    const { logout } = this.props;
+    const { logout, userProfile } = this.props;
```
burgwyn: Let's consolidate some of these ` = this.props` statements

src/Components/Header/Header.jsx
```diff
@@ -2,123 +2,128 @@ import React, { Component } from 'react';
...
+            <div className="usa-logo" id="logo">
+              <div className="usa-logo-text">
+                <Link to="/">
+                  <img src={logo} alt="TalentMAP logo" />
+                </Link>
+              </div>
             </div>
+            <DesktopNav
+              isLoggedIn={isLoggedIn}
+              showAccountDropdown={showAccountDropdown}
```
burgwyn: Is there a scenario where we would not be logged in AND show account dropdown?  I feel like we could pass fewer props and push some of that logic further down the stack.

src/Components/Header/Header.jsx
```diff
@@ -2,123 +2,128 @@ import React, { Component } from 'react';
...
 import GovBanner from './GovBanner/GovBanner';
-import AccountDropdown from '../AccountDropdown/AccountDropdown';
-import logo from '../../assets/logos/png/horizontal_color.png';
-import Inbox from './Inbox';
-import Notifications from './Notifications';
+import logo from '../../assets/logos/png/horizontal_color_thin.png';
+import ResultsSearchHeader from '../ResultsSearchHeader';
+import { isCurrentPath, isCurrentPathIn } from '../ProfileMenu/navigation';
+import { searchBarRoutes, searchBarRoutesForce } from './searchRoutes';
+import MobileNav from './MobileNav';
```
burgwyn: 👍 - good call splitting these up

src/Components/Header/Header.jsx
mjoyce91: Good point, I don't think so.

PR #1113, opened 2017-12-12
Migrate skill codes from strings to arrays

mjoyce91:
Migrate skill codes from string to arrays, based on changes in https://github.com/18F/State-TalentMAP-API/pull/263.

Also adds a new component to take a list of skill codes and render them in a comma separated list.


src/Components/ProfileDashboard/UserProfile/UserProfilePersonalInformation/UserProfilePersonalInformation.jsx
```diff
@@ -1,14 +1,15 @@
 import React from 'react';
 import { USER_PROFILE } from '../../../../Constants/PropTypes';
-import { NO_USER_SKILL_CODE, NO_BIRTHDAY } from '../../../../Constants/SystemMessages';
+import { NO_BIRTHDAY } from '../../../../Constants/SystemMessages';
```
burgwyn: 🚫🎂

mjoyce91: Just in case [Jason Bourne](https://en.wikipedia.org/wiki/The_Bourne_Identity_(2002_film)) uses TalentMAP

PR #1123, opened 2017-12-14
UI Testing

### LINKED ISSUE #1122: Setup dev environments to serve application at `/talentmap`
As mentioned in Slack, I think setting up the dev instance of the front end application to serve the application at `/talentmap` might be a good idea since it will better mirror the on-prem deployment. That way we'll easily be able to see bad image paths, etc during local development and will have fewer/no surprises with the on-prem.

Same for the back end as well, though I'm not clear on how that reverse proxy is setup on-prem. Is it the same path (`/talentmap`) but at a different host port?

burgwyn:
This closes #1122 and sets up the baseline branch for our UI testing efforts


src/Components/Header/Header.jsx
```diff
@@ -74,6 +74,8 @@ export class Header extends Component {
       shouldShowSearchBar, logout, userProfile,
     } = this.props;
 
+    const logo = getAssetPath('/assets/logos/png/horizontal_color.png');
```
mjoyce91: Current version of dev is using `horizontal_color_thin.png`, which was needed to make the header more narrow.

src/Containers/HomePage/HomePage.jsx
```diff
@@ -27,17 +27,18 @@ class HomePage extends Component {
...
       userProfile, toggleFavorite, toggleBid, bidList,
       userProfileFavoritePositionIsLoading,
       userProfileFavoritePositionHasErrored } = this.props;
     return (
       <div className="home content-container">
-        <Explore
-          filters={filters}
-          onRegionSubmit={this.submitRegion}
-        />
+        <div className="results results-search-bar-homepage">
```
mjoyce91: This whole div should be removed to reflect the search bar being moved to the header

src/Components/Footer/Footer.jsx
```diff
@@ -1,7 +1,12 @@
...
 import { Link } from 'react-router-dom';
+import { getAssetPath } from '../../utilities';
 import logo from '../../assets/logos/png/horizontal_color.png';
 
+const dosFooterLogo = getAssetPath('/assets/img/rsz_dos-seal.png');
+const tmFooterLogo = getAssetPath('/assets/logos/png/horizontal_color.png');
+
 const Footer = () => (
   <footer className="usa-footer usa-footer-medium tm-footer" role="contentinfo">
     <div className="usa-footer-primary-section tm-footer-nav">
```
mjoyce91: Compare the Footer against dev, something is thrown off:

screen shot 2017-12-14 at 11 51 24 am

src/utilities.js
```diff
@@ -2,6 +2,7 @@ import Scroll from 'react-scroll';
 import queryString from 'query-string';
 import distanceInWords from 'date-fns/distance_in_words';
 import { VALID_PARAMS } from './Constants/EndpointParams';
+// import paths from '../config/paths';
```
mjoyce91: Needed?

PR #1126, opened 2017-12-14
Limit CMSGov import to Spinner only

mjoyce91:
Only import the needed components from CMSGov (just the Spinner in our case) in order to avoid the USWDS version used by CMSGov (1.4.x) clashing with the one that we import (1.3.x).


src/sass/styles.scss
```diff
@@ -8,7 +8,7 @@
 @import '../../node_modules/typeface-roboto/index.css';
 
 // CMSGov Design System Layout
```
burgwyn: Let's update the comment

Issue #1116, opened 2017-12-12
UI - Create Bureau Information Component


mjoyce91:

burgwyn:
Closing #1132

Issue #1114, opened 2017-12-12
UI - Create and Style Bid Tracker Card with General Information


mjoyce91:

burgwyn:
Closing #1132

Issue #1049, opened 2017-12-04
UI - Style the step component


mjoyce91:

burgwyn:
Closing #1132

Issue #1048, opened 2017-12-04
UI - Map statuses to locations on the Step list


mjoyce91:

burgwyn:
Closing #1132

PR #1132, opened 2017-12-18
Bid Tracker - First Pass

mjoyce91:
First pass at the Bid Tracker. Implements core functionality to pull the user's bids and map them to the Bid Tracker. Doesn't include all functionality, but does include the following:

#1045 - Configure Steps module
#1048 - Map bid statuses to locations on the Steps list
#1049 - Style the Steps component
#1046 - Style supporting links and buttons
#1116 - Create bureau/org information component for bid tracker card
#1114 - Create and style bid tracker with general information
#1047 - Create list of steps from bid list

For now, the following are static:
- AO on card
- Bureau/Org on card
- Actions link
- Mail icon

Additionally, some status states appear the same as others because an overlay will appear over the tracker for those states (eg: In-panel and Approved are the same because an overlay will sit on top of the tracker if the status is Approved).

Should wait to merge until #1124 is complete.


src/Components/BidTracker/BidTrackerCardBottom/BidTrackerCardBottom.jsx
```diff
@@ -0,0 +1,20 @@
...
+import ExternalUserStatus from '../../ProfileDashboard/ExternalUserStatus';
+import OrganizationStamp from '../../OrganizationStamp';
+
+const BidTrackerCardBottom = () => (
+  <div className="usa-grid-full bid-tracker-card-bottom">
+    <div className="bid-tracker-ao-container">
+      <ExternalUserStatus showMail type="ao" name="Leah Shadtrach" />
+    </div>
+    <div className="bid-tracker-organization-container">
+      <OrganizationStamp
```
burgwyn: StaticWrapper

src/Components/ProfileMenu/ProfileMenu.jsx
```diff
@@ -17,6 +17,7 @@ const ProfileMenu = ({ isCDO }) => (
           search="?type=all"
           hidden={!isCDO}
         />
+        <NavLink title="Bid Tracker" link="/profile/bidtracker/" />
```
burgwyn: Reach out to Kristian and conform the order of these options

src/Components/BidTracker/BidTrackerCardBottom/BidTrackerCardBottom.jsx
mjoyce91: In a newer PR, I'm using real data, but the abbreviation is still static, so I'll put that in the wrapper.

burgwyn: Ok, let's move forward.  Thanks

jseppi:
Will need to re-open this against `dev` (or some new branch for all bid tracking features) once #1124 is merged to `staging`

PR #1136, opened 2017-12-19
Bid tracking alerts

mjoyce91:
Adds overlay alerts to the bid tracker steps. These are used for certain bid statuses where instead of displaying the tracker, a message or prompt is displayed to the user.

Covers #1115, #1117, #1118

Relies on #1132

Issue #1120, opened 2017-12-12
UI - Create Bid Tracker Notifications Alert and Supporting Redux Actions


mjoyce91:

burgwyn:
Closing #1143

PR #1143, opened 2017-12-20
Bid tracking notifications

mjoyce91:
Adds notifications and a "contact CDO button" to the bid tracker.

Covers #1120 

Relies on #1132 and #1136


src/Components/Alert/AlertAlt/AlertAlt.jsx
```diff
@@ -0,0 +1,50 @@
...
+      icon = 'check-circle';
+      break;
+    // default to info
+    default:
+      icon = 'info-circle';
+      break;
+  }
+  return (
+    // 'type' is injected into the class name
+    // type 'error' requires an ARIA role
```
burgwyn: 👍

src/Components/LinkButton/LinkButton.jsx
```diff
@@ -2,16 +2,28 @@ import React from 'react';
 import { Link } from 'react-router-dom';
 import PropTypes from 'prop-types';
 
-const LinkButton = ({ children, className, toLink, useDefaultClass }) => (
+const LinkButton = ({ children, className, toLink, useDefaultClass, useATag }) => (
```
burgwyn: Could you expand on `useATag`?  I understand the functionality, but what would be the use case?

mjoyce91: `mailTo` doesn't work with `Link` so we have to use a regular `<a>` tag.

jseppi: Could you just add some logic like `toLink.indexOf('mailto:') === 0` within the `LinkButton` component?

mjoyce91: We could, but external links can't be used in a `Link`. `LinkButton` could also be used as a way to navigate to external sites, a telephone number, etc, so it seems like it would be easier to set internal vs external explicitly, as a prop.

jseppi: ah, interesting. What about just checking to see if a protocol is specified at all? If it is, then it is "external."

If you still want to be explicit, maybe use something like `isExternal` instead of `useATag` since that kinda assumes some underlying knowledge of rendering that probably isn't necessary.

mjoyce91:
Using #1157

PR #1151, opened 2017-12-22
Bid Tracker - Pending Icon, Priority Wrappers, Bureau Data, Actions

mjoyce91:
Adds the following:
- "Pending" icon above the "Bid being evaluated" status
- Conditional rendering for Priority and Standby cards - work in progress, but adds templates that can be used once we have a way to determine whether a bid is priority
- Bureau data from the API instead of static
- Actions dropdown - static except for "Delete"
- Adds an InteractiveElement component to avoid redundancy in calling out the `eslint-disable-line` related to interactive elements not being properly picked up by CodeClimate (see #1152 for using this component across the application)

Relies on #1143


mjoyce91:
Moving to #1157.

PR #1156, opened 2017-12-26
Add profile menu collapse/expand functionality

### LINKED ISSUE #1153: Collapsable Profile Page Menu
Make the profile page menu collapsable so that it can be hidden and expanded.

mjoyce91:
Closes #1153

### DIFF
--- a/src/Components/ProfileDashboard/Notifications/NotificationItem/NotificationItem.test.jsx
+++ b/src/Components/ProfileDashboard/Notifications/NotificationItem/NotificationItem.test.jsx
@@ -1,6 +1,7 @@
 import React from 'react';
 import { shallow } from 'enzyme';
 import toJSON from 'enzyme-to-json';
+import { subMonths } from 'date-fns';
 import NotificationItem from './NotificationItem';
 
 describe('NotificationItemComponent', () => {
@@ -15,10 +16,14 @@ describe('NotificationItemComponent', () => {
   });
 
   it('matches snapshot', () => {
+    // Use subMonths so that snapshot doesn't go out of date every month.
+    // Using "2" means that our snapshot should always render "2 months ago"
+    // as its title.
+    const notificationTime = subMonths(new Date(), 2).toString();
     const wrapper = shallow(
       <NotificationItem
         content="content"
-        notificationTime="10-10-2017"
+        notificationTime={notificationTime}
       />,
     );
     expect(toJSON(wrapper)).toMatchSnapshot();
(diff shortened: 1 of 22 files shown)


src/Components/ProfileDashboard/Notifications/NotificationItem/__snapshots__/NotificationItem.test.jsx.snap
```diff
@@ -8,7 +8,7 @@ exports[`NotificationItemComponent matches snapshot 1`] = `
     className=""
     content="content"
     sideBySide={false}
-    title="2 months ago"
+    title="3 months ago"
```
jseppi: Uh oh! I think in `NotificationItem.test.jsx` we should construct and pass a time as the `notificationTime` prop so that this snapshot doesn't go out of date every month.

Something like:

```jsx
import { subMonths } from 'date-fns';

  // ...

  it('matches snapshot', () => {
    const notificationTime = subMonths(2, new Date());
    const wrapper = shallow(
      <NotificationItem
        content="content"
        notificationTime={notificationTime}
      />,
    );
    expect(toJSON(wrapper)).toMatchSnapshot();
  });
```

mjoyce91: Nice catch! Just updated.

src/Components/ProfileMenu/NavLink/NavLink.jsx
```diff
@@ -142,14 +147,19 @@ NavLink.propTypes = {
   // We still render hidden NavLinks so that we don't break NavLinksContainer.
   // They're simply returned as empty divs.
   hidden: PropTypes.bool,
+  isExpanded: PROFILE_MENU_SECTION_EXPANDED,
```
jseppi: calling this `isExpanded` is a little inconsistent with how we use `isX` for booleans. Maybe instead something like `expandedSection`?

src/Components/ProfileDashboard/Notifications/NotificationItem/NotificationItem.test.jsx
```diff
@@ -15,10 +16,14 @@ describe('NotificationItemComponent', () => {
   });
 
   it('matches snapshot', () => {
+    // Use subMonths so that snapshot doesn't go out of date every month.
```
jseppi: 👍

PR #1161, opened 2017-12-27
Add react-responsive module

mjoyce91:
Adds the `react-responsive` module and builds off of #1159 to use conditional rendering of components based on window width.


src/Components/MediaQuery/breakpoints.js
```diff
@@ -0,0 +1,11 @@
+// should match responsive breakpoints in _variables.scss
```
burgwyn: Move to constants

src/Components/Header/DesktopNav/DesktopNav.jsx
```diff
@@ -6,72 +6,75 @@ import Inbox from '../Inbox';
...
+import MediaQuery from '../../MediaQuery';
 
 const DesktopNav = ({ isLoggedIn, shouldShowSearchBar,
 userProfile, logout, toggleSearchVisibility }) => (
-  <div className="navigation-container">
-    <div className="nav-link-container header-nav-desktop desktop-nav-only">
-      {
-        isLoggedIn &&
-        <div className={`header-nav-link-container ${shouldShowSearchBar ? 'is-highlighted' : 'is-not-highlighted'}`}>
+  <MediaQuery widthType="min" breakpoint="screenMdMin">
```
burgwyn: Would we want this component living one level up?

mjoyce91: Do you mean wrapping `<DesktopNav>` with a `<MediaQuery>` from the `Header` file instead of in the `DesktopNav` file? I was actually trying to decide which to do - it's not like these components are ever used elsewhere, but I guess I can see it being more intuitive to handle the responsiveness from one component instead of two.

burgwyn: Yes, let's try in the `Header` to do
```
<MediaQuery small>
  <Mobile Nav />
</MediaQuery>
<MediaQuery large>
  <DesktopNav />
</MediaQuery>
```

src/sass/_variables.scss
```diff
@@ -41,6 +41,7 @@ $editor-headline-background-focus: #f3f3f3;
 $editor-headline-color: #888;
 
 // responsive breakpoints
+// should match responsive breakpoints in Constants/breakpoints.js
```
burgwyn: 👍

Issue #1172, opened 2017-12-28
Specify component props in tests


jseppi:
I noticed while running tests locally that several of them report that required props are not in the component tests. Would be good to clear up those warnings so that output is easier to read. It _might_ even be nice if the tests failed if props do not validate, but I don't know how to do that off the top of my head.

example (there are several others with similar warnings):

```sh
 PASS  src/Components/BidTracker/BidTrackerCardList/BidTrackerCardList.test.jsx
  ● Console

    console.error node_modules/fbjs/lib/warning.js:35
      Warning: Failed prop type: The prop `acceptBid` is marked as required in `BidTrackerCardList`, but its value is `undefined`.
          in BidTrackerCardList (at BidTrackerCardList.test.jsx:12)
    console.error node_modules/fbjs/lib/warning.js:35
      Warning: Failed prop type: The prop `declineBid` is marked as required in `BidTrackerCardList`, but its value is `undefined`.
          in BidTrackerCardList (at BidTrackerCardList.test.jsx:12)
    console.error node_modules/fbjs/lib/warning.js:35
      Warning: Failed prop type: The prop `acceptBid` is marked as required in `BidTrackerCard`, but its value is `undefined`.
          in BidTrackerCard (at BidTrackerCardList.jsx:10)
    console.error node_modules/fbjs/lib/warning.js:35
      Warning: Failed prop type: The prop `declineBid` is marked as required in `BidTrackerCard`, but its value is `undefined`.
          in BidTrackerCard (at BidTrackerCardList.jsx:10)
```

PR #1157, opened 2017-12-27
Bid Tracker - Contacts, Alert text, Rescheduled Prop, Tests/Comments

mjoyce91:
Addresses the following:

- New contacts on the Bid Tracker cards
- Tests and code comments
- #1138 - UI modifications for Rescheduled step
- #1139 - Notifications UI Changes on Bid Declined from Bureau and Rescheduled
- #1147 - UI - "taking so long?" & "priority bid?" tooltip/links
- Priority bid updates - relies on https://github.com/18F/State-TalentMAP-API/pull/287
- Rescheduled bid updates (new tooltip) - relies on https://github.com/18F/State-TalentMAP-API/pull/287

Using this as the main PR for pushing Bid Tracker related features into dev (closed two prior PRs).

PR #1175, opened 2017-12-29
Add setupTestFrameworkScriptFile to wrap errors [WIP]

mjoyce91:
Here's a way we could address #1172. This overrides the global `console.error` function used by Jest, and performs a regex for "Invalid prop" or "Failed prop" and throws an error if matched. The first instance of an invalid or failed prop will throw the `Error`, so only the first message within a failing test will be logged (i.e., if there are actually three failing prop types within a test, you'll only actually see the first).

Any non-matching error passed through `console.error` is placed inside a `console.log`, but that's only because of an issue I was running into where I was getting an empty `console.error` message. So I use [chalk](https://www.npmjs.com/package/chalk) so that the text is at least red.

This will actually fail tests on dev (which is good!), so if we like this approach, we can merge after #1157 is merged since it shouldn't have any proptype issues.


mjoyce91:
Another question is whether we'd want _any_ errors to throw an `Error`.

mjoyce91:
I did, that seemed to result the issue I described 😕

PR #1183, opened 2018-01-02
Add function to format date and apply across all uses of dates

mjoyce91:
Adds a function to convert dates as returned by the API to a custom format. Existing dates have been updated to use this new function.

### DIFF
--- a/src/utilities.test.js
+++ b/src/utilities.test.js
@@ -14,6 +14,7 @@ import { validStateEmail,
          existsInNestedObject,
          removeDuplicates,
          getTimeDistanceInWords,
+         formatDate,
        } from './utilities';
 
 describe('local storage', () => {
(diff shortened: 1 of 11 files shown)


mjoyce91:
`new Date()` accounts for timezone, but the dates we're using aren't timezone-bound. Need to update to ignore timezone.

src/utilities.test.js
```diff
@@ -224,3 +225,32 @@ describe('distanceInWords', () => {
     expect(timeDistanceInWords).toContain(' ago');
   });
 });
+
+describe('formatDate', () => {
+  it('returns a properly formatted date', () => {
+    // how we expect the date from the API
+    const unformattedDate = '2017-01-15';
+    // converted date
+    const formattedDate = formatDate(unformattedDate);
```
burgwyn: You're using addOneDay=true by default

mjoyce91: The tests pass locally that way though. When a date such as "2017-01-01" is provided (like we'd see it in the API), it gets converted to "Sat Dec 31 2016 19:00:00 GMT-0500 (EST)" for us on the East Coast. So assuming we want every one to see the same date, regardless of where they're located, we'd need to ignore the timezone.

mjoyce91: Trying to figure out the best way to do that.

mjoyce91:
Looks like this answer fixed timezone issues - https://github.com/date-fns/date-fns/issues/489#issuecomment-302316888

PR #1185, opened 2018-01-03
Add icon to Bid Tracker for bid preparing

mjoyce91:
Use new version of rc-steps which includes our `tailContent` PR. Use `tailContent` prop to pass `BidPreparingIcon` component.


mjoyce91:
Need to fix tooltip classes, will reopen.

PR #1208, opened 2018-01-08
Add Skill Code multiselect to home page

mjoyce91:
Adds a multiselect dropdown to the Homepage where users can select multiple skill codes to search for positions by.

### DIFF
--- a/src/Containers/Home/Home.jsx
+++ b/src/Containers/Home/Home.jsx
@@ -46,11 +46,12 @@ class Home extends Component {
       homePagePositionsHasErrored, homePagePositionsIsLoading,
       userProfile, toggleFavorite, toggleBid,
       userProfileFavoritePositionIsLoading, bidList,
-      userProfileFavoritePositionHasErrored } = this.props;
+      userProfileFavoritePositionHasErrored, filtersIsLoading } = this.props;
     return (
       <HomePage
         onNavigateTo={onNavigateTo}
         filters={items.filters}
+        filtersIsLoading={filtersIsLoading}
         homePagePositions={homePagePositions}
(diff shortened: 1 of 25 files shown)


src/Containers/Home/Home.jsx
```diff
@@ -46,11 +46,12 @@ class Home extends Component {
       homePagePositionsHasErrored, homePagePositionsIsLoading,
       userProfile, toggleFavorite, toggleBid,
       userProfileFavoritePositionIsLoading, bidList,
-      userProfileFavoritePositionHasErrored } = this.props;
+      userProfileFavoritePositionHasErrored, filtersIsLoading } = this.props;
```
burgwyn: Is there an instance where we have a <plural>IsLoading?

src/Containers/Home/Home.jsx
```diff
@@ -97,12 +99,13 @@ Home.defaultProps = {
...
 };
 
 const mapStateToProps = state => ({
   items: state.filters,
   hasErrored: state.filtersHasErrored,
-  isLoading: state.filtersIsLoading,
+  filtersIsLoading: state.filtersIsLoading,
   homePagePositions: state.homePagePositions,
   homePagePositionsHasErrored: state.homePagePositionsHasErrored,
   homePagePositionsIsLoading: state.homePagePositionsIsLoading,
```
burgwyn: nvm, we're doing <plural>IsLoading

src/Containers/Home/Home.jsx
mjoyce91: Yes, when the home page first loads and it has to fetch the filters. This gets passed to the `Select` component which renders a little spinner on the side.

PR #1218, opened 2018-01-09
Wait to render bid tracker until profile is finishing loading

### LINKED ISSUE #1209: UI Dev - Hide Bid Tracker and Show Spinner While Profile is Loading
Hide the Bid Tracker and show the loading spinner while it is loading.

mjoyce91:
Closes #1209.

### DIFF
--- a/src/Components/BidTracker/BidTracker.jsx
+++ b/src/Components/BidTracker/BidTracker.jsx
@@ -8,42 +8,47 @@ import ContactCDOButton from './ContactCDOButton';
 
 const BidTracker = ({ bidList, bidListIsLoading, acceptBid, declineBid,
 notifications, notificationsIsLoading, markBidTrackerNotification, userProfile,
-userProfileIsLoading }) => (
-  <div className="bid-tracker-page">
-    <NotificationsSection
-      notifications={notifications}
-      notificationsIsLoading={notificationsIsLoading}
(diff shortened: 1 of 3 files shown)


src/Components/BidTracker/BidTracker.jsx
```diff
@@ -8,42 +8,47 @@ import ContactCDOButton from './ContactCDOButton';
...
-      <div className="usa-width-one-half bid-tracker-cdo-email-container">
-        <div className="bid-tracker-cdo-email">
-          {
-            userProfile.cdo && !userProfileIsLoading &&
-            <ContactCDOButton email={userProfile.cdo.email} />
-          }
+        <div className="usa-width-one-half bid-tracker-cdo-email-container">
+          <div className="bid-tracker-cdo-email">
+            {
+              userProfile.cdo && !userProfileIsLoading &&
```
jseppi: should this just be hidden when `isLoading` as well?

mjoyce91: We could, but it doesn't rely on the bid data, as it's just a button to email your CDO. And since we display at the top, I think it's fine to render as soon as the profile has loaded.

jseppi: Alrighty, that seems fine 👍

Issue #998, opened 2017-11-21
Usability - Search Filter Ordering - Iteration


krisdelacruz:
As a Bidder
I want the filters to be ordered by Skill Code, grade and post
so that I can search efficiently for positions

## Findings
Users are filtering by Skill code, grade, and post

## Recommendation
Order filters by skill code, grade, and post

## User Research
The design team will continue to iterate on the filters. Phase 2 testing will reflect this research.

## Acceptance Criteria
Skill Code, Grade and Post are the first three filters in the search filters container

krisdelacruz:
This story is closely tired with #980 
 
We will have our participates ready for the new year

PR #1245, opened 2018-01-11
Reorder multi-select search filters

mjoyce91:
Addresses #998 by re-ordering the search filters on the results page. Also adds a test to ensure that the filters are in fact passed in the correct order, and centralizes the props in this test.


src/Components/SearchFilters/SearchFiltersContainer/SearchFiltersContainer.test.jsx
```diff
@@ -6,35 +6,56 @@ import SearchFiltersContainer from './SearchFiltersContainer';
...
+    },
+    {
+      item: { title: 'region', description: 'region', selectionRef: 'ref5' },
+      data: [{ isSelected: false, long_description: 'test', short_description: 'test' }],
+    },
   ];
 
   const accordion = ACCORDION_SELECTION;
 
+  const props = {
```
burgwyn: 👍

PR #1256, opened 2018-01-15
Add AuthorizedWrapper to conditionally render the Glossary component

### LINKED ISSUE #1254: Glossary doesn't load when user is logged out and then logs in
The glossary only loads if the session started with the user already logged in.

mjoyce91:
Closes #1254 by making use of an `AuthorizedWrapper` to conditionally render a component, the Glossary in this case, based on the result of `isAuthorized()`. We could apply this wrapper to other components throughout the application as well.

### DIFF
--- a/src/Containers/AuthorizedWrapper/AuthorizedWrapper.jsx
+++ b/src/Containers/AuthorizedWrapper/AuthorizedWrapper.jsx
@@ -1,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AuthorizedWrapper = ({ isAuthorized, children }) => {
+  const hasAuthorization = isAuthorized();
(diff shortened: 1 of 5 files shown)


src/Containers/AuthorizedWrapper/AuthorizedWrapper.jsx
```diff
@@ -0,0 +1,18 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+const AuthorizedWrapper = ({ isAuthorized, children }) => {
+  const hasAuthorization = isAuthorized();
+  return (
+    <div className="authorized-wrapper">
+      { hasAuthorization && children }
```
burgwyn: I think with React 16 we can use fragments here

PR #1262, opened 2018-01-16
Override placeholder color for accessibility

mjoyce91:
Default gray placeholder text for the Multi-select fails accessibility tests. This changes the text to a darker gray.


src/sass/_select.scss
```diff
@@ -0,0 +1,7 @@
+// scss-lint:disable SelectorFormat
+
```
burgwyn: Can you add a quick comment to indicate which component this overrides?

PR #1271, opened 2018-01-18
Redesign cards on the homepage

mjoyce91:
Redesigns the cards on the homepage based on recent mockups. There are still some small content changes that have yet to be made. We also use a static Time of Estimated Departure.

Another more noticeable change I made here is the gutter width. USWDS default max-width for the grid is 1040px, so I bumped that up to 1200px to more accurately reflect the the wireframes without going too far into full responsiveness.

I also updated the user dropdown to use the user's first name instead of their username.

**#1270 should be merged first or the dashboard will break!!**

### DIFF
--- a/src/Components/CondensedCardData/CondensedCardData.jsx
+++ b/src/Components/CondensedCardData/CondensedCardData.jsx
@@ -1,27 +1,34 @@
 import React from 'react';
-import { Link } from 'react-router-dom';
 import { POSITION_DETAILS } from '../../Constants/PropTypes';
-import { NO_POSITION_TITLE, NO_POST, NO_SKILL, NO_GRADE } from '../../Constants/SystemMessages';
+import { NO_POST, NO_SKILL } from '../../Constants/SystemMessages';
+import LanguageList from '../LanguageList';
+import StaticDevContent from '../StaticDevContent';
+import CondensedCardDataPoint from './CondensedCardDataPoint';
 
 const CondensedCardData = ({ position }) => (
   <div className="usa-grid-full condensed-card-data">
-    <div>
-      <strong>
-        <Link to={`/details/${position.position_number}`}>{ position.title || NO_POSITION_TITLE }</Link>
-      </strong>
-    </div>
-    <div>
-      <strong>Location: </strong>
-      { position.post ? position.post.location : NO_POST }
-    </div>
-    <div>
-      <strong>Skill Code: </strong>
-      { position.skill || NO_SKILL }
-    </div>
-    <div>
-      <strong>Grade: </strong>
-      { position.grade || NO_GRADE }
-    </div>
+    <CondensedCardDataPoint
+      title="Post"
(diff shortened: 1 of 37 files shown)


src/Components/CondensedCardData/CondensedCardData.jsx
```diff
@@ -1,27 +1,34 @@
...
+      title="Skill Code"
+      content={position.skill || NO_SKILL}
+      hasFixedTitleWidth
+    />
+    <CondensedCardDataPoint
+      title="Language"
+      content={<LanguageList languages={position.languages} />}
+      hasFixedTitleWidth
+    />
+    <StaticDevContent>
```
burgwyn: 👍

src/Components/CondensedCardData/CondensedCardDataPoint/CondensedCardDataPoint.test.jsx
```diff
@@ -0,0 +1,51 @@
...
+
+  it('applies the correct class when hasFixedTitleWidth is true', () => {
+    const wrapper = shallow(
+      <CondensedCardDataPoint
+        {...props}
+        hasFixedTitleWidth
+      />,
+    );
+    expect(wrapper.find('.condensed-card-data-title').exists()).toBe(true);
+    expect(wrapper.find('.title-fixed-width').exists()).toBe(true);
```
burgwyn: 👍

src/Components/ResultsCondensedCardTop/ResultsCondensedCardTop.jsx
```diff
@@ -1,59 +1,38 @@
 import React from 'react';
-import PropTypes from 'prop-types';
-import FontAwesome from 'react-fontawesome';
-import ResultsNewFlag from '../ResultsNewFlag';
-import Favorite from '../Favorite/Favorite';
-import { NO_LAST_UPDATED_DATE } from '../../Constants/SystemMessages';
-import { POSITION_DETAILS, FAVORITE_POSITIONS_ARRAY } from '../../Constants/PropTypes';
-import { formatDate } from '../../utilities';
+import { POSITION_DETAILS } from '../../Constants/PropTypes';
```
burgwyn: Let's remove the comments

src/Constants/DefaultProps.js
```diff
@@ -7,8 +7,8 @@ export const DEFAULT_HOME_PAGE_POSITIONS = {
 
 export const DEFAULT_USER_PROFILE = {
   user: {
-    username: '...', // show '...' when loading
-    first_name: '',
+    username: '...',
```
burgwyn: Should we move this to `''`?

mjoyce91: I've been using `'...'` as an easy way to display that it's loading

burgwyn: That's fine

PR #1279, opened 2018-01-22
New Comparison Page

mjoyce91:
Increases comparison limit to 5 and applies new styles to the Comparison page.


burgwyn:
`Error: Truffler timed out (30000ms)`
`Error: Phantom Process died`

What dark arts are you practicing @mjoyce91?!

mjoyce91:
Unfortunately there doesn't seem to be any consistency with these errors, as I just pushed an empty commit to retry the tests, and they passed 🤔

PR #1280, opened 2018-01-23
Bidder Portfolio Additional Details

mjoyce91:
Adds an additional details section to the Bidder Portfolio grid view. ~~This is a work in progress while @afrasier works on adding more fields to the `/client/:id/` response.~~

I approached this component slightly differently in that I don't use redux to maintain state, since the loading of each additional details section is a one-off request. An alternative could be to push new responses to an array stored in redux and use that as a cache for any instance where the user closes and re-opens an additional details section. Though I think I'd prefer for that data to get re-loaded in case there were any changes throughout the session.

Relies on https://github.com/18F/State-TalentMAP-API/pull/313


src/Components/BidderPortfolio/BidderPortfolioGridItemAdditional/ItemList/ItemList.test.jsx
```diff
@@ -0,0 +1,26 @@
+import React from 'react';
+import { shallow } from 'enzyme';
+import toJSON from 'enzyme-to-json';
+import ItemList from './ItemList';
+
+describe('ItemListingComponent', () => {
```
burgwyn: ItemList

src/Components/ProfileDashboard/PositionInformation/PositionInformation.jsx
```diff
@@ -8,8 +8,15 @@ import StartEnd from './StartEnd';
 import StaticDevContent from '../../StaticDevContent';
 
 const PositionInformation = ({ assignment }) => {
-  const assignmentStartDate = formatDate(assignment.start_date);
-  const assignmentEndDate = formatDate(assignment.estimated_end_date);
+  const assignmentStartDate = assignment.start_date ? formatDate(assignment.start_date) : false;
```
burgwyn: Can we use a static string like `N/A` for these empty dates?

src/utilities.test.js
```diff
@@ -326,3 +328,25 @@ describe('numbersToPercentString', () => {
...
+
+describe('formatBidTitle', () => {
+  it('can format a bid title', () => {
+    const bid = {
+      position: {
+        position_number: '0AA',
+        title: 'Title',
+      },
+    };
+    expect(formatBidTitle(bid)).toBeDefined();
```
burgwyn: We should test these a little further.  Can you make a `const` with the expected result?

src/Components/BidderPortfolio/BidderPortfolioGridItem/BidderPortfolioGridItem.jsx
```diff
@@ -1,43 +1,73 @@
...
-  <div className="usa-grid-full current-user bidder-portfolio-grid-item">
-    <div className="usa-width-one-fourth grid-item-section">
-      <div className="checkbox-container">
-        <CheckBox id={`checkbox-${userProfile.id}`} label="Select this user" value={false} labelSrOnly small />
+class BidderPortfolioGridItem extends Component {
+  constructor(props) {
+    super(props);
+    this.expandSection = this.expandSection.bind(this);
+    this.state = {
+      expanded: { value: false },
```
jseppi: any reason to nest the `value` here?
ie, why not move it up a "level" like:

```js
this.state = { expanded: false };
```

src/Components/BidderPortfolio/BidderPortfolioGridItem/BidderPortfolioGridItem.jsx
```diff
@@ -1,43 +1,73 @@
...
+          <div
+            className={`usa-width-one-fourth grid-item-section
+              current-user-personal-information-grid-container`}
+          >
+            <UserProfilePersonalInformation userProfile={userProfile} />
+          </div>
+          <div className="usa-width-one-fourth grid-item-section">
+            <UserProfileBidInformation
+              /* Object [0] should be the most recent. If undefined, use 0. */
+              draft={userProfile.bid_statistics[0] ? userProfile.bid_statistics[0].draft : 0}
```
jseppi: Maybe move the existence checking and defaults up to a variable for readability

```js
const latestBidStatistics = userProfile.bid_statistics[0] || { draft: 0, submitted: 0 };
```
```jsx
<UserProfileBidInformation
  draft={latestBuildStatistics.draft}
  submitted={latestBuildStatistics.submitted}
/>

```

src/Components/BidderPortfolio/BidderPortfolioGridItemAdditional/AdditionalContainer/AdditionalContainer.jsx
```diff
@@ -0,0 +1,60 @@
...
+      isLoading: { value: false },
+      hasErrored: { value: false },
+    };
+  }
+  componentWillMount() {
+    this.fetchClient();
+  }
+  // Since displaying each set of additional details is a one-off request,
+  // we don't use redux to maintain state and simply store it in the component's
+  // state.
```
jseppi: I think this is fine if it the data are not used elsewhere. If they do need to get used elsewhere for some other feature then the state can be lifted at that point.

src/Components/BidderPortfolio/BidderPortfolioGridItemAdditional/AdditionalContainer/AdditionalContainer.jsx
```diff
@@ -0,0 +1,60 @@
...
+  fetchClient() {
+    const { clientId, getClient } = this.props;
+    const clientPromise = getClient(clientId);
+    this.setState({ isLoading: { value: true } });
+    this.setState({ hasErrored: { value: false } });
```
jseppi: same question with the nested `value` as mentioned above

src/Components/BidderPortfolio/BidderPortfolioGridItem/BidderPortfolioGridItem.test.jsx
```diff
@@ -10,6 +10,12 @@ describe('BidderPortfolioGridItemComponent', () => {
     expect(wrapper).toBeDefined();
   });
 
+  it('can set expanded state to true with the expandSection function', () => {
```
jseppi: What about also testing for the effects of the `expanded` state being false/true?

src/Components/ProfileDashboard/PositionInformation/PositionInformation.jsx
mjoyce91: I didn't because then we'd still be displaying them within the `StartEnd` component and you'd end up with some thing like "N/A -> N/A"

src/Components/BidderPortfolio/BidderPortfolioGridItemAdditional/AdditionalView/AdditionalView.jsx
```diff
@@ -0,0 +1,85 @@
...
+import PropTypes from 'prop-types';
+import PositionInformation from '../../../ProfileDashboard/PositionInformation';
+import ItemList from '../ItemList';
+import Spinner from '../../../Spinner';
+import AlertAlt from '../../../Alert/AlertAlt';
+import { CLIENT_BY_ID } from '../../../../Constants/PropTypes';
+import { formatBidTitle, formatWaiverTitle } from '../../../../utilities';
+import { DRAFT_PROP, SUBMITTED_PROP } from '../../../../Constants/BidData';
+
+// eslint-disable-next-line
```
jseppi: what's this for?

I don't necessarily mind disabling rules here and there, as long as clarity is maintained, but it would be great if the specific rule was disabled instead of a blanket disabling

```js
// eslint-disable-next-line whatever-the-rule-is
```

src/Components/ProfileDashboard/PositionInformation/PositionInformation.jsx
burgwyn: Makes sense to me

src/Constants/PropTypes.js
```diff
@@ -463,3 +463,21 @@ export const BID_STATISTICS_OBJECT = PropTypes.shape({
...
+export const CLIENT_BY_ID = PropTypes.shape({
+  id: PropTypes.number,
+  current_assignment: ASSIGNMENT_OBJECT,
+  skills: USER_SKILL_CODE_ARRAY,
+  grade: PropTypes.string,
+  is_cdo: PropTypes.bool,
+  primary_nationality: PropTypes.string,
+  secondary_nationality: PropTypes.string,
+  bid_statistics: PropTypes.arrayOf(BID_STATISTICS_OBJECT),
+  user: PropTypes.shape({
```
jseppi: This user `shape` is used a couple times in this file. Might want to consider creating a variable for reuse

src/Components/BidderPortfolio/BidderPortfolioGridItem/BidderPortfolioGridItem.jsx
mjoyce91: I've noticed in the past that re-renders don't occur immediately if a string is directly modified versus modifying an object.

jseppi: huh, that's weird. I thought that was essentially what `setState`'s purpose was, ie, to signal state had changed and force a re-render. It looks like you're using `setState` everywhere.

src/Components/BidderPortfolio/BidderPortfolioGridItemAdditional/AdditionalView/AdditionalView.jsx
mjoyce91: Oops, must have used this during local testing. Will remove and include `hasErrored` prop type.

src/Components/BidderPortfolio/BidderPortfolioGridItem/BidderPortfolioGridItem.jsx
mjoyce91: Yeah, it seems to work fine in the case of this component. I think I had that issue with a text input, and so I've just been applying the same thing to other components.

jseppi: Something with redux-form perhaps? I've found that it can be...finicky.

Issue #1243, opened 2018-01-11
UI - Indicate position's bid cycle


burgwyn:
As a bidder
I need to know what cycle a position belongs to 
So that I know if the position is eligible for bidding

## Acceptance Criteria
A visual indicator of the bid cycle for each position

## Notes
Deploy to UI test site
Rough -- will need some design work later

PR #1286, opened 2018-01-23
Add bid cycle to results cards

mjoyce91:
Addresses #1243 by adding the Bid Cycle to the search results cards. Also adds the posted date.


burgwyn:
I think we can address #1 now.  I think #2 should be part of larger plan.

mjoyce91:
I'm not positive, but I think the `bid_statistics` object that we pull the name from is directly related to the bid cycle (i.e. they have the same `id`). @afrasier ?

And yes I agree with 2 - I think that will need to be part of a larger build out of bid cycle toggling.

afrasier:
Yeah, `bid_statistics` is via the route `/api/v1/bidcycle/{id}/statistics/`; it's generated on the fly for whichever bidcycle you specify.

PR #1301, opened 2018-01-26
Draft bid submission and deletion from Bid Tracker

mjoyce91:
Adds an Overlay alert to the Bid Tracker for draft bids where you have the option to submit or delete the bid, as well as see basic information about the position.

I wasn't able to find the exact issue number associated with this task, but I'll add it as soon as I find it.


mjoyce91:
This also includes a CSS fix to bid acceptance/decline buttons.

PR #1329, opened 2018-01-29
React-loadable

mjoyce91:
Integrates `react-loadable` and adds a common set of functions to easily wrap a normal component as a loadable component. For example, instead of directly using the `TextEditor` component, we could create the following the wrapper:

```jsx
...
import createLoader from '../../Loadable';

export const path = () => import('../TextEditor');

const TextEditor = createLoader({ path });

const TextEditorLoadable = ({ ...rest }) => (
  <TextEditor {...rest} />
...
);
```

`createLoader` takes what ever dynamic import we want and passes it to `react-loadable` with some params that we've defined, such as a whether it `shouldPreload`, and if so, how long it should wait to do so.

The only thing I'm running into issues with is using dynamic imports with Jest. Ideally, I'd like to test the loadable component at different states (pre-loading, loading, completed, error)... but Jest throws the following error: `System.import is not defined`. So for some reason Jest is not compiling the same way as the application, or maybe needs some global available. I've tried incorporating different babel plugins, but Jest still has issues. However, I don't think it should be a blocker ATM.


mjoyce91:
About to remove unused dependencies, but these were some I tried with no luck:

```js
"babel-plugin-dynamic-import-webpack": "^1.0.2",
"babel-plugin-syntax-dynamic-import": "^6.18.0",
"babel-plugin-system-import": "^1.1.5",
"babel-plugin-system-import-transformer": "^3.1.0",
```

PR #1350, opened 2018-01-31
Sprint 14

burgwyn:
This branch includes a lot of smaller fixes for CodeClimate as well as some other bugs


.babelrc
```diff
@@ -0,0 +1,6 @@
+{
+  "plugins": [
+    "syntax-dynamic-import",
```
jseppi: Hrm, the doc page for this says:

> This plugin only allows Babel to parse this syntax. If you want to transform it then see dynamic-import-webpack or dynamic-import-node.

Do we need to add `dynamic-import-webpack` as wel?

mjoyce91: I'll look into this. On the [readme](https://github.com/airbnb/babel-plugin-dynamic-import-webpack#babel-plugin-dynamic-import-webpack) it says `Note that Webpack 2 has gotten import() after this code was written.`. So maybe that's why it works in the app and not in the tests...?

PR #1359, opened 2018-02-01
Remove version

mjoyce91:
Remove version for now, since importing from `package.json` falls outside of `src` and throws an error during compliation.


jseppi:
Maybe another way to approach would be to read the `package.json` during webpack builds and use the [Define Plugin](https://webpack.js.org/plugins/define-plugin/) to put the value into the bundle.

mjoyce91:
https://github.com/18F/State-TalentMAP/issues/1362

Issue #1167, opened 2017-12-28
UI - Integrate and style Homepage banner components


mjoyce91:

burgwyn:
Closing

PR #1363, opened 2018-02-02
New search header

mjoyce91:
Covers #1165, #1166 and #1167 by adding new filters to the home page search bar.

Some notes about behavior:

- MediaQuery is used to render the original search component in condensed views. However, we share state so that the search term is maintained between component mounts.
- The user's skill code and grade are used to pre-populate the filters. The current selections are also saved to a shared state to be maintained between screen width changes. However, on any route change, they are reset to the defaults (an empty search string; user's skills and grade; no bureau selected) 
- Not available on the Results page since it already has a set of filters

### DIFF
--- a/src/Components/ResultsMultiSearchHeader/ResultsMultiSearchHeader.jsx
+++ b/src/Components/ResultsMultiSearchHeader/ResultsMultiSearchHeader.jsx
@@ -81,7 +81,7 @@ class ResultsMultiSearchHeader extends Component {
       this.setState({ defaultGrade, gradeWasUpdated: true });
     }
 
-    // set grade to correct state
+    // set bureau to correct state
     if (!bureauWasUpdated && defaultBureau) {
       this.setState({ defaultBureau, bureauWasUpdated: true });
     }
--- a/src/Components/ResultsMultiSearchHeader/ResultsMultiSearchHeader.jsx
+++ b/src/Components/ResultsMultiSearchHeader/ResultsMultiSearchHeader.jsx
@@ -2,9 +2,16 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import FontAwesome from 'react-fontawesome';
 import { FILTER_ITEMS_ARRAY, USER_PROFILE, EMPTY_FUNCTION } from '../../Constants/PropTypes';
+import { ENDPOINT_PARAMS } from '../../Constants/EndpointParams';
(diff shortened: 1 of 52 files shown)


src/Components/ResultsMultiSearchHeader/ResultsMultiSearchHeader.jsx
```diff
@@ -0,0 +1,231 @@
...
+    if (!qWasUpdated && defaultQuery) {
+      this.setState({ q: defaultQuery, qWasUpdated: true });
+    }
+
+    // set grade to correct state
+    if (!gradeWasUpdated && defaultGrade) {
+      this.setState({ defaultGrade, gradeWasUpdated: true });
+    }
+
+    // set grade to correct state
```
mjoyce91: need to update comment: grade -> bureau

src/Containers/HomePage/HomePage.jsx
```diff
@@ -27,18 +26,14 @@ class HomePage extends Component {
   }
 
   render() {
+    /* eslint-disable */
```
mjoyce91: Need to double check why I added this `eslint-disable`

jseppi: @mjoyce91 what are you doing working so late? :)

burgwyn: 🌜🦉

src/Components/ResultsMultiSearchHeader/ResultsMultiSearchHeader.jsx
```diff
@@ -0,0 +1,231 @@
...
+    const mappedGrades = grades && grades.data ?
+      grades.data.slice().map(g => ({ ...g, value: g.code, text: g.code })) : [];
+
+    // format bureaus
+    const bureaus = filters.find(f => f.item && f.item.description === 'region');
+    const mappedBureaus = bureaus && bureaus.data ?
+      bureaus.data.slice().map(g => ({ ...g, value: g.code, text: g.short_description })) : [];
+
+    // set the default skills
+    // eslint-disable-next-line camelcase
```
burgwyn: Remove

mjoyce91: Ah yeah, since I moved those weird variable names to constants, I don't need this any more 👍

PR #1384, opened 2018-02-08
Add dot-prop

mjoyce91:
Adds the dot-prop library and creates a utility that can be used instead of traversing object properties. Applies this new utility to a few components.


src/Components/BidTracker/ActionsDropdown/ActionsDropdown.jsx
```diff
@@ -60,11 +70,9 @@ export class ActionsDropdown extends Component {
             <InteractiveElement
               type="div"
               role="link"
-              className={`${dropdownSegmentClass} ${disableDelete ? 'disabled' : ''}`}
```
jseppi: Are these changes to the dropdown supposed to be in this same PR? They look fine, but was just wondering if there was a separate issue/story for them.

mjoyce91: Yes, I should have mentioned it, but I'm hoping this helps with some of the Code Climate complexity issues - so I also abstracted out some logic that might help with that.

PR #1392, opened 2018-02-09
Homepage 1358

burgwyn:
Creating a generic homepage row and replicating the Highlighted Positions row as an example


burgwyn:
Ignore the CodeClimate issues (I initially targeted `dev`)

src/Components/HomePagePositionsList/HomePagePositionsList.jsx
```diff
@@ -0,0 +1,80 @@
...
+    userProfileFavoritePositionIsLoading, userProfileFavoritePositionHasErrored, toggleBid,
+    bidList }) => {
+  // create an initial array with x groups of 3
+  // because our grid is in thirds
+  const numberOfRows = maxLength / 3;
+  const rows = Array(numberOfRows).fill(null);
+
+  // Form positions into component and add them to array
+  for (let i = 0; i < numberOfRows; i += 1) {
+    // load positions into an array of 3
```
mjoyce91: We could also use the `.usa-end-row` for every third item and keep them in the same `grid`, but this works too 👍

src/Components/HomePagePositionsList/HomePagePositionsList.jsx
```diff
@@ -0,0 +1,80 @@
...
+  favorites: [],
+  isLoading: false,
+};
+
+const HomePagePositionsList = ({ maxLength, positions, toggleFavorite, favorites, isLoading,
+    userProfileFavoritePositionIsLoading, userProfileFavoritePositionHasErrored, toggleBid,
+    bidList }) => {
+  // create an initial array with x groups of 3
+  // because our grid is in thirds
+  const numberOfRows = maxLength / 3;
```
mjoyce91: It might help to have this `3` be a prop so that can change it. I think some of our responsive views will only be 2 wide, so it would be helpful to do that with a prop.

src/Components/HomePagePositionsList/HomePagePositionsList.jsx
```diff
@@ -0,0 +1,80 @@
...
+
+  // Form positions into component and add them to array
+  for (let i = 0; i < numberOfRows; i += 1) {
+    // load positions into an array of 3
+    const positionList = Array(3).fill(null);
+
+    // copy prop list of positions, copy
+    // and push postions into a list
+    const positionsCopy = positions;
+    positionsCopy.slice((i * 3), ((i + 1) * 3)).forEach((p, j) => {
```
mjoyce91: Same thing here - using a `.usa-end-row` might reduce the need for this logic.

src/Components/HomePagePositionsList/HomePagePositionsList.jsx
mjoyce91: But then we'd have to map numbers to uswds width classes, which might be a pain 😣

PR #1397, opened 2018-02-12
Homepage 1358

burgwyn:
Map skills and grade to homepage positions

### DIFF
--- a/src/Containers/Home/Home.jsx
+++ b/src/Containers/Home/Home.jsx
@@ -27,7 +27,8 @@ class Home extends Component {
       this.props.onNavigateTo(PUBLIC_ROOT);
     } else {
       this.getFilters();
-      this.props.homePagePositionsFetchData();
+      this.props.homePagePositionsFetchData(this.props.userProfile.skills,
+        this.props.userProfile.grade);
       this.props.bidListFetchData();
     }
   }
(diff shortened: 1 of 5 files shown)


src/Containers/Home/Home.jsx
```diff
@@ -27,7 +27,8 @@ class Home extends Component {
       this.props.onNavigateTo(PUBLIC_ROOT);
     } else {
       this.getFilters();
-      this.props.homePagePositionsFetchData();
+      this.props.homePagePositionsFetchData(this.props.userProfile.skills,
```
mjoyce91: This won't work if the home page is the first page to load. The actual `userProfile` doesn't exist until the profile is retrieved, and `componentWillMount` is called before then.

mjoyce91: We could move these `HomePagePositionSection`s into a connected component and do something like:

Homepage.jsx
```jsx
...
{ !userProfileIsLoading && <HomePagePositionsContainer userProfile={userProfile} /> }
...
```

And then move some of that logic from `Home.jsx` to the new `HomePagePositionsContainer.jsx`
```
componentWillMount() {
      this.props.homePagePositionsFetchData(this.props.userProfile.skills,
        this.props.userProfile.grade);
}
```

burgwyn:
Closing as this functionality was addressed in #1409

Issue #1368, opened 2018-02-06
Disable pa11y checks in CI process


burgwyn:
Our pa11y checks have failed regularly for the last few weeks and our efforts to resolve have not been fruitful.  Disable pa11y from the CI flow until a fix is identified.

burgwyn:
Note - disabled pa11y as a required check on these branches:
* dev
* staging
* master

burgwyn:
Closing

PR #1403, opened 2018-02-12
dev -> staging

mjoyce91:
Sprint 14 work. Includes the following

Features <span class="emoji-outer emoji-sizer"><span class="emoji-inner" style="background: url(chrome-extension://immhpnclomdloikkpcefncmfgjbkojmh/emoji-data/sheet_apple_32.png);background-position:12.044653349001175% 57.99059929494712%;background-size:5418.75% 5418.75%" data-codepoints="1f31f"></span></span>

- https://github.com/18F/State-TalentMAP/pull/1378 - Glossary CRUD
- https://github.com/18F/State-TalentMAP/pull/1363 - New search header
- https://github.com/18F/State-TalentMAP/pull/1392 - Homepage cards (will add)

Tweaks <span class="emoji-outer emoji-sizer"><span class="emoji-inner" style="background: url(chrome-extension://immhpnclomdloikkpcefncmfgjbkojmh/emoji-data/sheet_apple_32.png);background-position:55.99294947121034% 14.042303172737954%;background-size:5418.75% 5418.75%" data-codepoints="1f527"></span></span>

- https://github.com/18F/State-TalentMAP/pull/1329 - Add react-loadable <span class="emoji-outer emoji-sizer"><span class="emoji-inner" style="background: url(chrome-extension://immhpnclomdloikkpcefncmfgjbkojmh/emoji-data/sheet_apple_32.png);background-position:99.94124559341951% 53.99529964747356%;background-size:5418.75% 5418.75%" data-codepoints="2b50"></span></span>
- Code complexity warnings from Codeclimate
- https://github.com/18F/State-TalentMAP/pull/1343 - Typo
- https://github.com/18F/State-TalentMAP/pull/1384 - dot-prop
- Node proxy application for SAML integration

Bugs <span class="emoji-outer emoji-sizer"><span class="emoji-inner" style="background: url(chrome-extension://immhpnclomdloikkpcefncmfgjbkojmh/emoji-data/sheet_apple_32.png);background-position:26.02820211515864% 22.032902467685076%;background-size:5418.75% 5418.75%" data-codepoints="1f41e"></span></span>

- https://github.com/18F/State-TalentMAP/pull/1331 - Fix formatFraction function
- https://github.com/18F/State-TalentMAP/pull/1349 - `preventDefault`


jseppi:
I like this format for the root PR comment!

jseppi:
@mjoyce91: have you run `pa11y` on this locally? I'm thinking it might be good to start including those results with these PRs (as a comment/screenshot) since we've had to disable it due to CI issues (ref #1338).

mjoyce91:
@jseppi - I actually didn't realize they were disabled. I can start doing that for dev -> staging PRs.

jseppi:
tagged the wrong issue - fixed it above (https://github.com/18F/State-TalentMAP/issues/1368)

src/Components/ProfileMenu/ProfileMenuCollapsed/ProfileMenuCollapsed.jsx
```diff
@@ -14,6 +14,7 @@ const ProfileMenuCollapsed = ({ expand }) => (
     <NavLinksContainer>
       <NavLink iconName="user" link="/profile/dashboard/" />
       <NavLink iconName="pie-chart" link="/profile/statistics/" />
+      <NavLink iconName="book" link="/profile/glossaryeditor/" search="?type=all" />
```
jseppi: Do we have a story to restrict this to only users who have edit privileges?

src/Components/GlossaryEditor/GlossaryEditorCardBottom/GlossaryEditorCardBottom.jsx
```diff
@@ -0,0 +1,57 @@
...
+  const showWarningOrError = showEmptyWarning || showResponseError;
+
+  return (
+    <div className="usa-grid-full glossary-card-bottom-container">
+      <div className="glossary-warning-container">
+        {
+          showWarningOrError &&
+            <span className="usa-input-error-message" role="alert">
+              { showEmptyWarning && 'Title and definition cannot be blank.' }
+              { showResponseError && 'Error updating term.' }
```
jseppi: Could a more descriptive error be shown based on the API response? For instance, when a duplicate term is submitted and the request fails, it's not really clear to the user why it failed.

src/Components/GlossaryEditor/ArchiveIcon/ArchiveIcon.jsx
```diff
@@ -0,0 +1,52 @@
...
+    const { isArchived } = this.state;
+    // set initial values as if term is NOT archived
+    let icon = 'eye-slash';
+    let className = 'term-not-archived';
+    let text = 'Archive this term';
+
+    if (isArchived) {
+      icon = 'eye';
+      className = 'term-archived';
+      text = 'Un-archive this term';
```
jseppi: Minor bug: when you try to archive a term, but it fails, the `state` still has `isArchived: true` so it looks like it worked (though the error message is still shown)

jseppi: Also the length of this text ends up wrapping to the next line, which looks a bit off:

screen shot 2018-02-13 at 4 12 20 pm

src/Components/GlossaryEditor/GlossaryEditorContainer/GlossaryEditorContainer.jsx
```diff
@@ -0,0 +1,44 @@
...
+        submitGlossaryTerm={submitGlossaryTerm}
+        terms={glossaryItems}
+        availableLetters={availableLetters}
+        glossaryPatchHasErrored={glossaryPatchHasErrored}
+        glossaryPatchSuccess={glossaryPatchSuccess}
+      />
+      {
+        noResults &&
+        <div className="usa-width-two-thirds">
+          <Alert title="There are no Glossary terms." messages={[{ body: 'Try adding the first one.' }]} />
```
jseppi: When there are no search results, an alert should probably also be shown:

screen shot 2018-02-13 at 4 20 58 pm

src/Components/GlossaryEditor/ArchiveIcon/ArchiveIcon.jsx
jseppi: I guess that will also happen depending on how long the editor's name is, too. Probably best not to worry about that until visual design for this is completed.

src/Components/HomePagePositionsList/HomePagePositionsList.jsx
```diff
@@ -1,62 +1,82 @@
...
-  });
-  return (
-    <div className={`usa-grid-full condensed-card-highlighted ${isLoading ? 'results-loading' : ''}`}>
-      <div className="usa-grid-full">
+      );
+    });
+
+    // load into rows array
+    rows.push(
+      <div key={shortid.generate()} className="usa-grid-full">
```
jseppi: could make a "combo" key here based on the positionList items so that re-renders don't happen unnecessarily. something like:

```js
function genKey(positionListItems) {
   // returns a hyphenated string of position ids, eg '11-14-22'
   const key = positionListItems.map(p => p.id).join('-');
   return key;
}
```

```jsx
<div key={genKey(positionList.slice(0,3))} className="usa-grid-full">
```

src/Constants/PropTypes.js
```diff
@@ -92,6 +92,8 @@ export const POST_SEARCH_RESULTS = PropTypes.shape({
 export const HOME_PAGE_POSITIONS = PropTypes.shape({
   isNew: POSITION_DETAILS_ARRAY,
   isHighlighted: POSITION_DETAILS_ARRAY,
+  isSkillCode: POSITION_DETAILS_ARRAY,
+  isGrade: POSITION_DETAILS_ARRAY,
```
jseppi: A little weird that these are all name as booleans, but are actually arrays. How about something like 

```
isNew -> newPositions
isHighligted -> highlightedPositions
isSkillCode -> userSkillCodePositions
isGrade -> userGradePositions
```

src/Containers/GlossaryEditor/GlossaryEditor.jsx
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  componentWillMount() {
+    const { fetchGlossary } = this.props;
+    fetchGlossary();
+  }
+
+  submitGlossaryTerm(term) {
+    const { submitGlossaryTerm } = this.props;
+    // TODO - create action and integrate with API once its implemented
```
jseppi: Does this need an issue? I thought the API did support this. Maybe the timing was just off during development.

https://github.com/18F/State-TalentMAP-API/blob/dev/talentmap_api/glossary/views/glossary.py#L39

src/Components/TextEditor/TextEditor/TextEditor.jsx
```diff
@@ -1,9 +1,9 @@
 import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import Editor, { createEditorStateWithText } from 'draft-js-plugins-editor';
-import TextEditorSubmit from '../TextEditorSubmit';
-import { EMPTY_FUNCTION } from '../../Constants/PropTypes';
-import InteractiveElement from '../InteractiveElement';
+import TextEditorSubmit from '../../TextEditorSubmit';
+import { EMPTY_FUNCTION } from '../../../Constants/PropTypes';
+// import InteractiveElement from '../../InteractiveElement';
```
jseppi: remove?

mjoyce91:
@jseppi - the index.js file in the TextEditor folder points to TextEditorLoadable. With this approach, we can later decide to make a component loadable without going back and updating component names.

src/Containers/GlossaryEditor/GlossaryEditor.jsx
mjoyce91: Looks like an old comment - will delete

src/utilities.js
```diff
@@ -262,11 +263,16 @@ export const returnObjectsWherePropMatches = (sourceArray = [], compareArray = [
 // Convert a numerator and a denominator to a percentage. Pass "inverse === true" if you want
 // the inverse, i.e. the remainder.
 export const numbersToPercentString = (numerator, denominator, precision = 3) => {
-  const formatFraction = fraction => parseFloat(fraction).toPrecision(precision) * 100;
+  const formatFraction = fraction =>
```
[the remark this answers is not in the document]
mjoyce91: This looks cool - https://github.com/adamwdraper/Numeral-js

src/Components/BidderPortfolio/TopNav/Navigation/NavigationItem/NavigationItem.jsx
```diff
@@ -7,21 +7,29 @@ import { isCurrentParam } from '../../../../ProfileMenu/navigation';
 
 export const NavigationItem = ({ title, numerator, denominator, link, location }) => {
   const isUnderlined = isCurrentParam(link, location.search, 'type');
+  // numerator can be a zero, but not null or undefined
+  const numeratorIsNotUndefinedOrNull = numerator !== null && numerator !== undefined;
+  const showFraction = numeratorIsNotUndefinedOrNull && denominator;
```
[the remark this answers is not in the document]
mjoyce91: Yeah, I think not displaying a fraction at all if the denominator is zero makes sense.

src/Components/ProfileMenu/ProfileMenuCollapsed/ProfileMenuCollapsed.jsx
mjoyce91: @afrasier - could we return something on the `/profile` route like `is_glossary_editor: true/false`?

src/Components/GlossaryEditor/GlossaryEditorCard/GlossaryEditorCard.jsx
```diff
@@ -0,0 +1,169 @@
...
+    this.updateDefinition = this.updateDefinition.bind(this);
+    this.cancel = this.cancel.bind(this);
+    this.submitDefinition = this.submitDefinition.bind(this);
+    this.state = {
+      editorHidden: true,
+      newTitle: null,
+      newTitleWasEdited: false,
+      newDefinition: null,
+      newDefinitionWasEdited: false,
+      displayZeroLengthAlert: false,
```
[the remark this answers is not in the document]
mjoyce91: I did that as an easy way to say "if false, display nothing", but if it exists then it's an object noting whether the title or definition is empty `{ title: true/false, definition: true/false }`. However, what I decided to do was just return a catch-all for either - `const showEmptyWarning = emptyTitleWarning || emptyDefinitionWarning;`, so that ended up being pointless.

I'll go back and just make this a boolean for both instances since that is a bit confusing.

src/Components/GlossaryEditor/GlossaryEditorCardBottom/GlossaryEditorCardBottom.jsx
mjoyce91: Good point, I'll hook into the API response and return it to the user.

src/Components/GlossaryEditor/GlossaryEditorCard/GlossaryEditorCard.jsx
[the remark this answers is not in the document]
mjoyce91: https://github.com/18F/State-TalentMAP/issues/1426

src/Components/HomePagePositionsList/HomePagePositionsList.jsx
mjoyce91: Oh neat, good idea

mjoyce91: So thinking about this, I'm not sure we could actually guarantee uniqueness. Technically there's nothing stopping two rows from having the same positions in the same order, however unlikely it may be.

src/Constants/PropTypes.js
mjoyce91: https://github.com/18F/State-TalentMAP/issues/1429

jseppi:
Oh, cool, that works 👍 Thanks for pointing it out. Maybe a comment in `Components/TextEditor/index.js` to call some attention to it would be good :)

src/Components/HomePagePositionsList/HomePagePositionsList.jsx
jseppi: You're right -- and that would be a worse situation. `shortid` it is!

src/utilities.js
```diff
@@ -262,11 +263,16 @@ export const returnObjectsWherePropMatches = (sourceArray = [], compareArray = [
 // Convert a numerator and a denominator to a percentage. Pass "inverse === true" if you want
 // the inverse, i.e. the remainder.
```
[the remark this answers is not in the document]
mjoyce91: https://github.com/18F/State-TalentMAP/pull/1430

src/Components/GlossaryEditor/GlossaryEditorContainer/GlossaryEditorContainer.jsx
mjoyce91: https://github.com/18F/State-TalentMAP/pull/1430

src/Containers/GlossaryEditor/GlossaryEditor.jsx
mjoyce91: https://github.com/18F/State-TalentMAP/pull/1430

src/Components/TextEditor/TextEditor/TextEditor.jsx
mjoyce91: https://github.com/18F/State-TalentMAP/pull/1430

mjoyce91:
Included TextEditor comment in https://github.com/18F/State-TalentMAP/pull/1430

src/utilities.js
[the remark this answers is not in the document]
jseppi: oh, yeah, I came across that as well. Seems like it would be generally useful for us, and it's not very big.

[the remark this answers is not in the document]
jseppi: could also use it for adding commas to make large numbers more readable, eg `Showing 1-10 of 6,000` positions`

src/Components/ProfileMenu/ProfileMenuCollapsed/ProfileMenuCollapsed.jsx
afrasier: `/api/v1/permission/user/` will show all permission a user has, you could check for the presence of "glossary_editor"  in their groups. Unless it is much more preferable to have the is_glossary_editor flag; but I feel like as permissions go out into the future we may need to determine a better way for this information to be transferred to the front end.

jseppi: I agree; adding flags for each permission seems a bit gnarly to me.

PR #1410, opened 2018-02-13
Node Proxy

burgwyn:
Simple node app to proxy the SAML2 metadata endpoint


src/server.js
```diff
@@ -0,0 +1,30 @@
+const express = require('express');
+const path = require('path');
+
+// define the API root url
+const API_ROOT = process.env.API_ROOT || 'http://localhost:8000';
+
+// define the url consuming the API, provide local dev value as default
+const API_URL = process.env.API_URL || `${API_ROOT}/api/v1`;
```
mjoyce91: Is this variable ever used?

src/server.js
```diff
@@ -0,0 +1,30 @@
...
+const port = process.env.PORT || 3000;
+
+const app = express();
+
+// middleware for static assets
+app.use(PUBLIC_URL, express.static(path.join(__dirname, '../build')));
+
+// saml2 metadata
+app.get(`${PUBLIC_URL}metadata`, (request, response) => {
+  response.redirect(`${API_ROOT}/saml2/metadata`);
```
mjoyce91: Maybe you meant to use URL here? Should the saml2 route have the `/api/v1` prefix?

burgwyn:
The SAML and API pieces have separate routes, which was why I created to the two variables.  We're not proxying any of the API functions currently, so I removed it.

src/server.js
```diff
@@ -0,0 +1,30 @@
...
+// application port
+const port = process.env.PORT || 3000;
+
+const app = express();
+
+// middleware for static assets
+app.use(PUBLIC_URL, express.static(path.join(__dirname, '../build')));
+
+// saml2 metadata
+app.get(`${PUBLIC_URL}metadata`, (request, response) => {
```
mjoyce91: The API usually requires trailing slashes. Is that needed here as well?

PR #1419, opened 2018-02-13
Add tests for proxy server

mjoyce91:
Separates the app and server from `server.js` and exports them so they can be tested. Tests server routes for `200`/success and `304`/redirect responses.

Targeted at `sprint-15` branch.


burgwyn:
I think we need to make the source file directory, in this case `../build`, an environment variable

mjoyce91:
It seemed like this was working for me when I first tried it, but now I can only get the base HTML file to load. Does this code have anything to do with it?

```js
app.get('*', (request, response) => {
  response.sendFile(path.resolve(STATIC_PATH, 'index.html'));
});
```

This seems to be returning all files as the index.html. Notice how the JS file is actually just `index.html`:

screen shot 2018-02-15 at 9 09 27 am

burgwyn:
Closing this and picking up the work on another branch

PR #1438, opened 2018-02-15
Add redirect for OBC urls

mjoyce91:


src/server.js
```diff
@@ -7,6 +7,10 @@ const API_ROOT = process.env.API_ROOT || 'http://localhost:8000';
 // define the prefix for the application
 const PUBLIC_URL = process.env.PUBLIC_URL || '/talentmap/';
 
+// define the OBC root url
+// example: https://www.obcurl.gov
+const OBC_URL = process.env.OBC_URL;
```
jseppi: can you add documentation for this env var to the README?

jseppi: Actually I don't think we have `PUBLIC_URL` documented there either, so it would be good to have that spelled out as well.

Happy to have that done as a separate issue/PR if you'd prefer.

burgwyn:
This hits the intended spirit.  We have 2 types of OBC content: countries and posts.  Can we have two routes, `/obc/country/:id` and `/obc/post/:id`

burgwyn:
I'll handle the documentation in a separate PR.

Issue #1421, opened 2018-02-13
Error with npm run build


mjoyce91:
When building the application, the following error appears, but the application still runs fine.

```
Skipping static resource "/<app_path_here>/build/static/js/main.6ec69b1f.js" (3.94 MB) - max size is 2.1 MB
Failed to compile.

static/js/main.6ec69b1f.js from UglifyJs
Unexpected token name «i», expected punc «;» [./~/dot-prop/index.js:8,0][static/js/main.6ec69b1f.js:65373,10]
```

burgwyn:
Looks like `dot-prop` is [not made to be transpiled](https://github.com/sindresorhus/dot-prop/issues/44)

PR #1448, opened 2018-02-16
Webpack fix

burgwyn:
This addresses #1421


jseppi:
Does this change also need to happen in https://github.com/18F/State-TalentMAP/blob/dev/config/webpack.config.dev.js#L167?

burgwyn:
No, this is specific to the Uglify process

PR #1450, opened 2018-02-16
Home Page Personalization Fallback

mjoyce91:
Adds fallback queries in the case that the user does not have skill and/or grade codes. Skill positions get replaced with Favorited positions, and Grade positions gets replaced with Recently Posted positions.


mjoyce91:
The easiest way to test this is to remove grade and skill from one of your users.

PR #1451, opened 2018-02-16
Add numeral.js

mjoyce91:
Use [numeral.js](https://github.com/adamwdraper/Numeral-js) to format percentages.


src/utilities.test.js
```diff
@@ -311,19 +311,18 @@ describe('returnObjectsWherePropMatches', () => {
 describe('numbersToPercentString', () => {
   let numerator = 2;
   let denominator = 10;
-  let precision = 2;
+  const format = '0.00%';
```
jseppi: maybe move this down into the second `it` block since it's only used there

Issue #1314, opened 2018-01-29
Fix "method_complexity" issue in src/Components/BidList/BidList.jsx


jseppi:
Function `BidList` has a Cognitive Complexity of 8 (exceeds 5 allowed). Consider refactoring.

https://codeclimate.com/github/18F/State-TalentMAP/src/Components/BidList/BidList.jsx#issue_5a6f3a942b12380001000449

PR #1452, opened 2018-02-16
Remove the standalone Bid List page

mjoyce91:
Removes the standalone Bid List page and related components/containers/links, as we've migrated away from this page to the Bid Tracker.

Closes https://github.com/18F/State-TalentMAP/issues/1314


burgwyn:
We need this against `release-candidate-1` too

mjoyce91:
@burgwyn - what's the best way to merge release-candidate-specific work? What if we tracked work in three buckets: 1. Release-candidate work, 2. Work for both, 3. Non-release-candidate work. Then we wouldn't need to create duplicate PRs for each individual feature.

burgwyn:
I was looking to tag issues with product vs release candidate (or both).  I think we can save feature branches, re-base and re-PR.  It doesn’t solve the multiple PR problem, but it would let us isolate at the issue level.

mjoyce91:
It looks like I accidentally removed the Bid Tracker here as well. We'll want to remove that for `release-candidate-1`, but not our main branch.

PR #1453, opened 2018-02-18
SAML Proxy

### LINKED ISSUE #1442: Proxy SAML Endpoints
Frontend wrapper for backend implementation.  Urls here https://github.com/knaperek/djangosaml2/blob/master/djangosaml2/urls.py

burgwyn:
Closes #1442


mjoyce91:
Did you see my previous screenshot? I believe this is returning `index.html` for any request that's not defined in the server. I.e., js, css, etc are returned with the contents of `index.html`.

mjoyce91:
What if we did something like:

```js
// could export/import these from Router.jsx 
app.get(['/', '/login*', '/results*', '/profile*', ...], (request, response) => {
  response.sendFile(path.resolve(STATIC_PATH, 'index.html'));
});

app.get('*', (request, response) => {
  response.sendfile(STATIC_PATH + request.params[0]);
});
```

PR #1454, opened 2018-02-19
Export/import routes from react app to server

mjoyce91:
Exports the routes from React into the express server so that React routes can be rendered when navigated to by the express server.


burgwyn:
The old wildcard route seemed to handle these routes, but we'll need this in the long run.

PR #1467, opened 2018-02-21
Updating deployment document

burgwyn:
Adding steps, removing nvm

### DIFF
--- a/DEPLOY.md
+++ b/DEPLOY.md
@@ -8,7 +8,7 @@ Install core server dependencies
 
 ```bash
 sudo yum update
-sudo yum install httpd mod_ssl -y
+sudo yum install httpd mod_proxy mod_ssl -y
 sudo yum install openssl -y
 sudo yum install git -y
 ```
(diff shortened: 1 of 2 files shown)


DEPLOY.md
```diff
@@ -1,67 +1,103 @@
...
+Use `git` to clone the web app repository
 
 ```bash
 git clone https://github.com/18F/State-TalentMAP.git
+cd State-TalentMAP/
+```
+
+## Source environment variables
+
+Environment variables are documented [here](EXAMPLE_setup_environment.sh)
```
jseppi: I think this link is wrong since that file is over in the API repo: https://github.com/18F/State-TalentMAP-API/blob/dev/EXAMPLE_setup_environment.sh

jseppi: But sourcing those env vars isn't necessary for the front end app, right?

burgwyn: I'm trying to mirror the documentation on the API for consistency.  The `EXAMPLE_setup_environment.sh` lives on the front end on a different branch.  I'll correct.

burgwyn: Check that, the file is here https://github.com/18F/State-TalentMAP/blob/deploy-doc/EXAMPLE_setup_environment.sh

jseppi: Ah, cool. Thanks!

Might want to remove the lines about being parsed by python (https://github.com/18F/State-TalentMAP/blob/deploy-doc/EXAMPLE_setup_environment.sh#L3-L4), but looks good otherwise.

PR #1478, opened 2018-02-23
Replace dot-prop with local utility

mjoyce91:
IE11 has issues with `dot-prop`, so I replaced it with a local utility that performs the same job. It has the added benefit of returning the property value for `0`s.

### DIFF
--- a/config/webpack.config.prod.js
+++ b/config/webpack.config.prod.js
@@ -158,7 +158,6 @@ module.exports = {
       // Process JS with Babel.
       {
         test: /\.(js|jsx)$/,
-        exclude: /node_modules\/(?!dot-prop)\/*/,
         loader: require.resolve('babel-loader'),
 
       },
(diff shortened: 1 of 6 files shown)


burgwyn:
Can you review the `webpack` files and remove the `dot-prop` exemption?

mjoyce91:
Nice catch

burgwyn:
And remove from `package.json` and `yarn.lock`

config/webpack.config.prod.js
```diff
@@ -158,7 +158,6 @@ module.exports = {
       // Process JS with Babel.
       {
         test: /\.(js|jsx)$/,
```
burgwyn: Add `include: paths.appSrc,`

jseppi:
Apologies for recommending a library that ended up having these weird issues.

PR #1502, opened 2018-02-27
Update font versions

burgwyn:


package.json
```diff
@@ -38,8 +38,8 @@
     "redux-saga": "^0.15.6",
     "redux-thunk": "^2.2.0",
     "shortid": "^2.2.8",
-    "typeface-merriweather": "^0.0.45",
-    "typeface-roboto": "0.0.50",
+    "typeface-merriweather": "^0.0.54",
```
jseppi: do you need to also update `yarn.lock` to get the updated version there?

burgwyn: Good call.  I tried to do this in the browser during the deployment.

PR #1521, opened 2018-03-01
User Feedback

mjoyce91:
Adds a way for users to submit feedback. Relies on https://github.com/18F/State-TalentMAP-API/pull/360.

### DIFF
--- a/src/Components/Feedback/Feedback.jsx
+++ b/src/Components/Feedback/Feedback.jsx
@@ -1,0 +1,81 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import FontAwesome from 'react-fontawesome';
+import Form from '../Form';
+import FieldSet from '../FieldSet/FieldSet';
+import TextEditor from '../TextEditor';
+import CheckBox from '../CheckBox';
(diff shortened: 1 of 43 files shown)


src/Components/Feedback/Feedback.jsx
```diff
@@ -0,0 +1,82 @@
+import React from 'react';
```
burgwyn: You want to rename to `FeedbackForm` since we have a `Feedback` container?

src/Containers/FeedbackButton/FeedbackButton.jsx
```diff
@@ -0,0 +1,47 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import { connect } from 'react-redux';
+import { toggleFeedback } from '../../actions/showFeedback';
+import { EMPTY_FUNCTION } from '../../Constants/PropTypes';
+import FeedbackButton from '../../Components/FeedbackButton';
+
+class FeedbackContainer extends Component {
```
burgwyn: `FeedbackButton`

PR #1523, opened 2018-03-02
Permission Groups

mjoyce91:
Fetches the user's permission groups along with their profile and combines the two objects into one. Then uses this data to determine whether to display the Glossary Editor link in the profile menu.

### DIFF
--- a/src/Components/ProfilePage/ProfilePage.jsx
+++ b/src/Components/ProfilePage/ProfilePage.jsx
@@ -9,10 +9,11 @@ import BidStatistics from '../../Containers/BidStatistics';
 import GlossaryEditor from '../../Containers/GlossaryEditor';
 import ProfileMenu from '../ProfileMenu';
 import { USER_PROFILE } from '../../Constants/PropTypes';
+import { userHasPermissions } from '../../utilities';
 
 const ProfilePage = ({ user }) => (
   <div className="profile-page">
-    <ProfileMenu isCDO={user.is_cdo} />
+    <ProfileMenu isCDO={user.is_cdo} isGlossaryEditor={userHasPermissions(['glossary_editors'], user.permission_groups)} />
     <div className="usa-grid-full profile-content-container">
       <Switch>
         <Route path="/profile/dashboard" component={Dashboard} />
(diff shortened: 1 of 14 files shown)


src/Components/ProfilePage/ProfilePage.jsx
```diff
@@ -9,10 +9,11 @@ import BidStatistics from '../../Containers/BidStatistics';
 import GlossaryEditor from '../../Containers/GlossaryEditor';
 import ProfileMenu from '../ProfileMenu';
 import { USER_PROFILE } from '../../Constants/PropTypes';
+import { userHasPermissions } from '../../utilities';
 
 const ProfilePage = ({ user }) => (
   <div className="profile-page">
-    <ProfileMenu isCDO={user.is_cdo} />
+    <ProfileMenu isCDO={user.is_cdo} isGlossaryEditor={userHasPermissions(['glossary_editors'], user.permission_groups)} />
```
burgwyn: Let's list our permissions in `Constants`

src/utilities.js
```diff
@@ -314,3 +314,14 @@ export const formatIdSpacing = (id) => {
   // if id is not defined, return null
   return null;
 };
+
+// provide an array of permissions to check if they all exist in an array of user permissions
+export const userHasPermissions = (permissionsToCheck = [], userPermissions = []) => {
+  let hasPermissions = true;
```
burgwyn: Let's invert this.  User should be assumed to NOT have a permission.

mjoyce91: I found a more straight forward way to check:

```js
permissionsToCheck.every(val => userPermissions.indexOf(val) >= 0)
```

src/utilities.test.js
```diff
@@ -395,3 +396,42 @@ describe('formatIdSpacing', () => {
...
+    expect(userHasPermissions(permissionsToCheck, userPermissions)).toBe(false);
+  });
+
+  it('returns true if the user has more than the needed permissions', () => {
+    userPermissions.push('d', 'e');
+    expect(userHasPermissions(permissionsToCheck, userPermissions)).toBe(true);
+  });
+
+  it('returns false if userPermissions are not provided', () => {
+    userPermissions = [];
```
burgwyn: I almost feel like we should throw an exception in this case as it might be a sign that the user's session is stale

PR #1525, opened 2018-03-02
Branded Loading Spinner

burgwyn:
This may need some additional styling feeback


mjoyce91:
Looks good - there was actually one scenario where we need the spinner to be much smaller, and that's when we use `react-loadable` for the text editor. I added a `size` prop to the Spinner so that we can do this.

I think it would also help to load the SVGs into the bundle, as there's a small delay between the page's initial load and the time it takes to load all the SVGs. I tried using [`svg-inline-loader`](https://github.com/webpack-contrib/svg-inline-loader) but wasn't able to get it working.

If my changes look good to you, I can approve and we can merge.

PR #1571, opened 2018-03-08
Remove `ResultsViewBy` from `ResultsControls`

### LINKED ISSUE #1555: UI Dev - Remove "View By" Toggle
Remove the "View by" toggle on the Search Results page, as we will only have one card style to display.

Icemanbeta:
- Updated component snapshot

Resolves #1555


Icemanbeta:
The `ResultsViewBy` component is also used by `BidControls`. Maybe we should consider renaming it to a more vanilla name - `ViewBy` (or `ViewByFilter`, `ViewByControls`, or whatever makes sense lol)

mjoyce91:
@Icemanbeta good idea - that makes sense. Feel free to create an issue

PR #1581, opened 2018-03-09
- Updated all calls to `localStorage` -> `sessionStorage`

Icemanbeta:
- Updated all calls to `localStorage` -> `sessionStorage`
- Updated all relevant variable names  that use *localStorage*


Icemanbeta:
@burgwyn I should open a PR to get this feature into `dev` right?

src/Components/CompareCheck/CompareCheck.jsx
```diff
@@ -12,14 +12,14 @@ class CompareCheck extends Component {
...
+      sessionStorageKey: null,
       count: 0,
     };
   }
 
   componentWillMount() {
-    const localStorageKey = this.props.type;
-    this.setState({ localStorageKey });
+    const sessionStorageKey = this.props.type;
+    this.setState({ sessionStorageKey });
```
mjoyce91: Here's something I hadn't thought about with the Comparison feature: I wonder if we'd actually want to store the user's comparisons in local storage. I feel like it would be good UX for those comparisons to persist between sessions, since we don't store them on the API. We'd want to clear them between user sessions, however, so that User A doesn't automatically receive the comparison selections of previously-logged-in User B.

src/lib/check-auth.test.js
```diff
@@ -27,15 +27,15 @@ describe('check-auth', () => {
   });
 
   it('can return false when a token is not set in local storage', () => {
```
mjoyce91: local storage -> session storage

src/lib/check-auth.test.js
```diff
@@ -27,15 +27,15 @@ describe('check-auth', () => {
...
-    localStorage.clear();
+    sessionStorage.clear();
     expect(checkIndexAuthorization(store)).toBe(false);
-    localStorage.clear();
+    sessionStorage.clear();
   });
 
   it('can return false when a token is not set in local storage', () => {
```
mjoyce91: local storage -> session storage

src/utilities.test.js
```diff
@@ -27,41 +27,41 @@ import { validStateEmail,
 
 describe('local storage', () => {
```
mjoyce91: local storage -> session storage

src/Components/CompareCheck/CompareCheck.jsx
Icemanbeta: Actually I like that idea, I totally wasn't thinking when I made the global change.

Icemanbeta:
I'm so bad lol. Sorry about unit tests :innocent:

src/utilities.test.js
Icemanbeta: Reverted changes to compare components (which ended up reverting `utilities.js`)

Icemanbeta:
- Updated all relevant variable names  that use *localStorage*

burgwyn:
Build is failing because `jest-localstorage-mock` requires node v6.11.3 or greater and we're running 6.11.0 in CircleCI.

src/lib/check-auth.js
```diff
@@ -1,8 +1,8 @@
 import { setClient } from '../client/actions';
 
 function checkAuthorization(dispatch) {
-  // attempt to grab the token from localstorage
-  const storedToken = localStorage.getItem('token');
+  // attempt to grab the token from sessionStorage
+  const storedToken = sessionStorage.getItem('token');
```
burgwyn: @mjoyce91 - should we use the utility `fetchUserToken` here?

mjoyce91: Yeah, might as well use that here too.

burgwyn: Cool, I'll update

Icemanbeta:
@mjoyce91 We might be able to implement cross tab communication: https://truongtx.me/2014/06/16/cross-tab-communication-using-html5-dom-storage

Icemanbeta:
@burgwyn @mjoyce91 Maybe we don't need this anymore with the stale token interceptor being implemented

burgwyn:
Agreed, closing

PR #1582, opened 2018-03-11
- Workaround for redux dispatch function prop causing all instances o…

### LINKED ISSUE #1351: Favoriting and Adding to Bid List feel unresponsive
### Steps to reproduce the issue (please be as specific as possible)
Click on the ⭐️ icon to make a position a favorite (or not-favorite).
Click on Add to Bid List to add a position to your bid list.

### Expected behavior
The icon toggles to favorite/not-favorite state quickly.
The Add to Bid List button response quickly.

### Actual behavior
There is a noticeable delay in the icon state changing.

favorite-slow

### Suggestions
- What if we toggle the visual state _first_ before all the async actions are complete? Then, if the async actions fail (for any reason -- network error, too many in bid list, etc), we can revert the visual state. I'm not sure if that UX is great or not
- Show a spinner or some other loading state if the async actions take more than `THRESHOLD ms`

Icemanbeta:
- Workaround for redux dispatch function prop causing all instances of `Favorite` component to re-render by adding a `shouldComponentUpdate` hook.

### DIFF
--- a/src/Components/Favorite/Favorite.jsx
+++ b/src/Components/Favorite/Favorite.jsx
@@ -6,18 +6,34 @@ import { existsInArray } from '../../utilities';
 import InteractiveElement from '../InteractiveElement';
 
 class Favorite extends Component {
-
   constructor(props) {
     super(props);
     this.toggleSaved = this.toggleSaved.bind(this);
+
+    this.state = {
+      loading: props.isLoading,
+    };
   }
 
-  shouldComponentUpdate(nextProps) {
+  shouldComponentUpdate(nextProps, nextState) {
+    let isUpdate = true;
+
     const { compareArray, refKey } = nextProps;
     const oldState = this.getSavedState();
     const newState = existsInArray(refKey, compareArray);
 
-    return (oldState !== newState);
+    if (this.state.loading && !nextProps.isLoading) {
+      // Only update the loading state if current state.loading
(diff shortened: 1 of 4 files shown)


Icemanbeta:
Check https://github.com/18F/State-TalentMAP/issues/1351#issuecomment-372083753

mjoyce91:
This looks good. I think we should still implement some kind of UI, as you mention in https://github.com/18F/State-TalentMAP/issues/1351#issuecomment-372083753. I think this could either be a spinner, or we could simply change the state, and revert it should an AJAX error occur.

Icemanbeta:
If an ajax error occurs, the component wouldn't render the active flag I dont think because the component always rerenders based on the `compareArray`. 

Roger that on the loading UI I'll update this

Icemanbeta:
- Add loading spinner UI to `Favorite` - https://github.com/18F/State-TalentMAP/issues/1351#issuecomment-37247499

Icemanbeta:
- Tweaked `Favorite` container when child of `ResultCard` - https://github.com/18F/State-TalentMAP/issues/1351#issuecomment-372516644 😄

burgwyn:
@Icemanbeta - run `yarn test -u` - it will regenerate the snapshots with the latest changes.  More info [here](https://facebook.github.io/jest/docs/en/snapshot-testing.html)

PR #1583, opened 2018-03-13
Release Candidate - Profile Page - Saved Searches

mjoyce91:
Adds a saved searches section to the Profile Dashboard. I've created a connected wrapper that can be used to wrap other components that need Saved Search-related data. I also used the same filter mapping found on the Results page to create pills that show human-readable names for the Saved Search.


burgwyn:
Testing and it looks good.  My user had no saved searches so I created #1584

PR #1594, opened 2018-03-14
Shared configuration for `ProfileMenuCollapsed` and `ProfileMenuExpanded`

Icemanbeta:
- Shared configuration for `ProfileMenuCollapsed` and `ProfileMenuExpanded`
- Updated snapshot for `ProfileMenuCollapsed`
- Updated snapshot for `ProfileMenuExpanded`

### DIFF
--- a/src/Components/ProfileMenu/ProfileMenuExpanded/ProfileMenuExpanded.jsx
+++ b/src/Components/ProfileMenu/ProfileMenuExpanded/ProfileMenuExpanded.jsx
@@ -1,67 +1,116 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import FontAwesome from 'react-fontawesome';
-import { PROFILE_MENU_SECTION_EXPANDED_OBJECT } from '../../../Constants/DefaultProps';
-import { PROFILE_MENU_SECTION_EXPANDED } from '../../../Constants/PropTypes';
 import NavLinksContainer from '../NavLinksContainer';
 import NavLink from '../NavLink';
 
-const ProfileMenuExpanded = ({ isCDO, expandedSection, collapse, toggleMenuSection,
-isGlossaryEditor }) => (
-  <div className="usa-grid-full profile-menu">
-    <div className="menu-title">
-      <div className="menu-title-text">Menu</div>
-      <button className="unstyled-button" title="Collapse menu" onClick={collapse}>
-        <FontAwesome name="exchange" />
-      </button>
+import { PROFILE_MENU_SECTION_EXPANDED_OBJECT } from '../../../Constants/DefaultProps';
+import { PROFILE_MENU_SECTION_EXPANDED } from '../../../Constants/PropTypes';
+import { PROFILE_MENU } from '../../../Constants/Menu';
+
+function getProps(options, params = {}) {
+  const props = {
+    title: options.text,
+    iconName: options.icon,
+    link: options.route,
+    search: (options.params || ''),
+    hidden: (options.isCDO && !params.isCDO) ||
+            (options.isGlossaryEditor && !params.isGlossaryEditor),
+  };
+
+  if (options.toggleMenuSection || options.expandedSection) {
+    delete props.link;
+
+    if (options.toggleMenuSection) {
+      props.toggleMenuSection = params.toggleMenuSection;
+    }
+
+    if (options.expandedSection) {
+      props.expandedSection = params.expandedSection;
+    }
+  }
+
+  return props;
+}
+
+/*
+<NavLink
+  title="Profile"
+  iconName="user"
+  toggleMenuSection={toggleMenuSection}
+  expandedSection={expandedSection}
+>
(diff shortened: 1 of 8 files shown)


src/Components/ProfileMenu/ProfileMenuExpanded/ProfileMenuExpanded.jsx
```diff
@@ -1,67 +1,116 @@
...
+
+    if (options.expandedSection) {
+      props.expandedSection = params.expandedSection;
+    }
+  }
+
+  return props;
+}
+
+/*
```
mjoyce91: Is this comment block to display expected output, or was it left behind?

src/Constants/Menu.js
```diff
@@ -0,0 +1,119 @@
...
+    text: 'Glossary Editor',
+    icon: 'book',
+    route: '/profile/glossaryeditor/',
+    isGlossaryEditor: true,
+    params: {
+      type: 'all',
+    },
+  },
+  {
+    text: 'Inbox',
```
mjoyce91: Non-profile sections shouldn't have the "expand" arrow since they don't have any child links. See screenshots:

screen shot 2018-03-14 at 9 02 34 am

screen shot 2018-03-14 at 9 02 46 am

src/Constants/Menu.js
```diff
@@ -0,0 +1,119 @@
...
+ *    route?: string;
+ *    params?: { [key: string]: string; };
+ *    toggleMenuSection?: boolean;
+ *    expandedSection?: boolean;
+ *    isCDO?: boolean;
+ *    isGlossaryEditor?: boolean;
+ *    children?: Array<MenuItem>;
+ *  }
+ */
+function params(data) {
```
mjoyce91: what does this function do?

src/Constants/Menu.js
```diff
@@ -0,0 +1,119 @@
...
+    route: '/profile/contacts/',
+  },
+  {
+    text: 'Documents',
+    icon: 'file-text',
+    route: '/profile/documents/',
+  },
+]);
+
+export default PROFILE_MENU;
```
mjoyce91: Can you add a test for this file?

mjoyce91:
Also need to figure out why CircleCI doesn't like `Object.entries`. Node version?

src/Components/ProfileMenu/ProfileMenuExpanded/ProfileMenuExpanded.jsx
Icemanbeta: Oops yea. I commented out the original code and forgot to take it off. Wanted to make sure j switch to compare. Thanks for the reminder!

src/Constants/Menu.js
Icemanbeta: I didn't catch this. This must be related to me not using self closing and hoping react didn't actually do something other than look at inner content. I guess I'll have to use a conditional for the loop

src/Constants/Menu.js
Icemanbeta: So for the search prop I noticed it was a queryparam string. Thought it be nice to scale the config a little bit and let it be an object of key values instead if we wanted to add more. The function just converts an object to a query string like jquery params

src/Constants/Menu.js
Icemanbeta: Oh yea for sure

src/Constants/Menu.js
mjoyce91: Oh cool, that makes sense. Can you add a comment above the function with that description?

src/Constants/Menu.js
```diff
@@ -1,10 +1,39 @@
...
+ * url encoded query parameter string
+ *
+ * @param {object} data Object with keys and string values
+ * @returns {string} Valid queryparam string with delimiters
+ * @example
+ *
+ * params({ key1: 'value1', key2: 'value2' })
+ * // => '?key1=value1&key2=value2'
+ */
+function params(data) {
```
mjoyce91: Something I forgot is that we actually already use the `query-string` library to handle this elsewhere, such as in the Results page. Might save ourselves from additional testing to use `query-string` instead.

src/sass/styles.scss
```diff
@@ -8,7 +8,7 @@
 @import '../../node_modules/typeface-roboto/index.css';
 
 // Spinner from the CMSGov Design System Layout
-@import '../../node_modules/@cmsgov/design-system-core/dist/components/Spinner/Spinner.css';
+// @import '../../node_modules/@cmsgov/design-system-core/dist/components/Spinner/Spinner.css';
```
mjoyce91: I know this is probably left over from another PR, but let's uncomment this.

src/Constants/Menu.js
Icemanbeta: Omg I saw this package before I implemented this. I didn't even think to check `package.json` sounds good to me

PR #1611, opened 2018-03-16
Profile Initial Avatar

Icemanbeta:
- ~Added `react-avatar` to dependencies~
- Added `react-addons-shallow-compare` to peer dependencies
- Update `src/sass/sass-vars/variables.js` to be more modular
- Update unit tests for `AccountDropdown`, `ProfilePicture`
- Update snapshots


burgwyn:
`error react-avatar@2.5.1: The engine "node" is incompatible with this module. Expected version "8.x.x"`

src/Components/AccountDropdown/AccountDropdown.jsx
```diff
@@ -25,19 +26,19 @@ export class AccountDropdown extends Component {
   render() {
     const { shouldDisplayName, userProfile } = this.props;
     const displayName = userProfile ? userProfile.display_name : '...';
```
burgwyn: 👍

src/Components/AccountDropdown/AccountDropdown.jsx
```diff
@@ -25,19 +26,19 @@ export class AccountDropdown extends Component {
...
-    const avatar = getAssetPath('/assets/img/avatar.png');
+    const fullName = ([
```
burgwyn: 👍

burgwyn:
Hey, I don't think we can use `react-avatar` because it targets Node 8 and we're running 6.  That fact that it has some baked in support for social networks also gives me pause.  Can we make our own "avatar" component instead?

Icemanbeta:
@burgwyn yea no I was thinking the same thing, on the social network part. For the Node thing, I looked into it a bit but figure I try the build anyways.

I'm on it!

Icemanbeta:
@burgwyn Fixed the commit =D, I should stop trying to use packages im unfamilar with. It's definitely easier doing our own. I guess unless we needed the social networks. This way will be more scalable for responsiveness in the future too (`react-avatar` does injected styles)

src/Components/ProfilePicture/ProfilePicture.jsx
```diff
@@ -1,27 +0,0 @@
-import React from 'react';
```
burgwyn: Let's keep this component for now.  There's talk of using profile pictures in a future release.

src/Components/ProfileDashboard/ExternalUserStatus/__snapshots__/ExternalUserStatus.test.jsx.snap
```diff
@@ -40,7 +41,7 @@ exports[`ExternalUserStatusComponent matches snapshot 1`] = `
           <span
             className="cdo-name-text"
           >
-            John Doe
+            nodejs
```
burgwyn: This looks weird

Icemanbeta: @burgwyn so actually I thought about that and well I think I would scale it in to the `Avatar` component and have it accept a `src` prop. If its passed in, then we'd render the `img` tag instead. We shouldn't have two different components for this I think. Anyways, the update would have the same template as `ProfilePicture`. Actually I can merge it in now, but I think having 2 components is too much. 

Actually, background image over the avatar in an inline css style might be better.

Icemanbeta: Let me know your thoughts

src/sass/_dropdown.scss
```diff
@@ -24,6 +24,13 @@
   height: 20px;
   margin: 0 5px 0 0;
   width: 20px;
+
+  .tm-avatar {
```
mjoyce91: The avatar seems a _little_ big on the top nav. Could we shrink it by about 10-20%?

src/Components/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,57 @@
...
+  get value() {
+    const value = [
+      this.getInitial(this.props.firstName),
+      this.getInitial(this.props.lastName),
+    ];
+
+    return this.toText(value, '');
+  }
+
+  getInitial = value => (value || '').charAt(0);
```
mjoyce91: @burgwyn made a handy `initials` prop on the `/profile/` response. Might as well use that here.

src/Components/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,57 @@
...
+  toText = (text, delimiter = ' ') => text.join(delimiter).trim();
+
+  render() {
+    const alt = this.caption;
+    const text = this.value;
+    const props = {
+      className: this.props.className,
+      onClick: this.props.onClick,
+      role: 'img',
+      'aria-label': alt,
```
mjoyce91: :+1:

src/Components/AccountDropdown/AccountDropdown.test.jsx
```diff
@@ -11,19 +11,28 @@ import { AccountDropdown } from './AccountDropdown';
...
   const mockStore = configureStore(middlewares);
+  const props = {
+    userProfile: {
+      display_name: 'John Doe',
+      user: {
+        first_name: 'John',
+        last_name: 'Doe',
+      },
+    },
+  }
```
mjoyce91: Might want to double check your eslint setup - there's a few things it flagged on this file (no semicolon, and an extra space on line 30).

src/Components/AccountDropdown/AccountDropdown.test.jsx
```diff
@@ -11,19 +11,28 @@ import { AccountDropdown } from './AccountDropdown';
...
 
   it('is defined', () => {
     const accountDropdown = shallow(<AccountDropdown />);
     expect(accountDropdown).toBeDefined();
   });
 
   it('can take different props', () => {
-    const accountDropdown = shallow(<AccountDropdown userProfile={{ user: { display_name: 'test' } }} />);
+    const accountDropdown = shallow(<AccountDropdown  />);
```
mjoyce91: Needs `{...props}`

mjoyce91: Check out the rest of the tests in this file and make sure they all have `{...props}`.

src/Components/ProfileDashboard/ExternalUserStatus/__snapshots__/ExternalUserStatus.test.jsx.snap
burgwyn: I'm comfortable with this approach.  Delete the old component and we'll revisit it in the future.

Have a look at the `{name}` issue.

src/Components/Avatar/Avatar.jsx
burgwyn: Yeah, I think we added `initials` after you started this work, but let's use it.

src/Components/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,57 @@
...
+    return this.toText(value);
+  }
+
+  get value() {
+    const value = [
+      this.getInitial(this.props.firstName),
+      this.getInitial(this.props.lastName),
+    ];
+
+    return this.toText(value, '');
```
mjoyce91: Should we return a `?` if the values in `value` are undefined, just to be safe? That way we don't risk returning nothing if for some reason both of those values are undefined.

src/Components/Avatar/Avatar.test.jsx
```diff
@@ -0,0 +1,35 @@
...
+    expect(wrapper.prop('aria-label')).toBe(caption);
+  });
+
+  it('can pass an optional className prop', () => {
+    const wrapper = shallow(<Avatar {...props} />);
+    expect(wrapper.find(`.${props.className}`).exists()).toBe(true);
+  });
+
+  it('matches snapshot', () => {
+    const wrapper = shallow(<Avatar />);
```
mjoyce91: `{...props}`?

src/Components/Avatar/Avatar.jsx
mjoyce91: I also mention that because based on our current login approach, stale tokens still keep you "logged in" to the site, and you have to manually log out. If the initials circle doesn't render, there's no way to do that.

src/Components/AccountDropdown/AccountDropdown.jsx
```diff
@@ -48,14 +49,10 @@ export class AccountDropdown extends Component {
             <div>Signed in as</div>
```
mjoyce91: Check out the `dropdown__content` class. I feel like this container got pushed a little too far to the addition of the user's first name. It'll probably be a little bit tricky because the first name is always a different length depending on the currently logged in user.

screen shot 2018-03-19 at 10 19 38 am

src/Components/ProfileDashboard/ExternalUserStatus/__snapshots__/ExternalUserStatus.test.jsx.snap
Icemanbeta: @burgwyn Any ideas why unit tests passed despite this obvious issue with `{name}`? Even though eslint may have caught it

src/sass/_dropdown.scss
Icemanbeta: 10% Reduction (`45px`)
image

20% Reduction (`40px`)
image

Original Wireframe
image

I think if you look at how much white space is left above the avatar and below you'll see it got a bit bigger in the design.

Original Size: (`50px`)

I committed a 10% reduction, let me know what you think.

src/Components/AccountDropdown/AccountDropdown.test.jsx
Icemanbeta: The only eslint package I have is actually local to this app and I haven't done any tweaks 😭 

Oh you know what...I did notice I never get warnings and errors on test files...I don't think it's what I did though. I think it's what you did maybe. Cause the webpack config isn't watching test files when the dev server is running.

Is it supposed to lint the test files on `yarn test`?

src/sass/_dropdown.scss
mjoyce91: Looks good! Thanks

src/Components/AccountDropdown/AccountDropdown.test.jsx
mjoyce91: Hmm. I don't think `yarn test` cares - it will still run un-linted test files. I'm also using some Atom plugins that automatically fix any errors on save.

src/Components/Avatar/Avatar.jsx
Icemanbeta: I see where you're coming from and well...😅 
PR https://github.com/18F/State-TalentMAP/pull/1640

src/Components/Avatar/Avatar.test.jsx
Icemanbeta: Shrug, I thought I tested the props too much so I didn't add it for the snapshots, but I suppose it doesn't hurt. Just trying to be different lol - 3745f9cb1b478de4a4464b51c66b868aa0794433

src/Components/AccountDropdown/AccountDropdown.jsx
Icemanbeta: How's this? - 3745f9cb1b478de4a4464b51c66b868aa0794433

mjoyce91: That looks good! Thanks!

burgwyn:
This looks good.  @Icemanbeta - can you modify `Avatar.jsx` to use the initials on the user's profile?  I think that's the last thing we need to close this.

mjoyce91:
:confused:

screen shot 2018-03-22 at 9 47 34 am

burgwyn:
Yeah, prop isn't passed all the way down

src/Components/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,32 @@
...
+const defaultProps = {
+  initials: '',
+  firstName: '',
+  lastName: '',
+  className: '',
+  onClick: EMPTY_FUNCTION,
+};
+
+const Avatar = ({ initials, firstName, lastName, className, onClick }) => (
+  <InteractiveElement className={className} onClick={onClick} role="img" aria-label={`${firstName} ${lastName}`}>
```
mjoyce91: I actually don't think the avatar needs to be an `InteractiveElement`. The `.dropdown__trigger` container in the top nav already receives focus, so using `InteractiveElement` adds a redundant focus-able element. The same goes for when it's displayed on the Profile Dashboard - clicking it doesn't actually do anything, but it receives a `tabIndex` and `pointer: cursor`, which could be confusing to the user.

burgwyn:
@Icemanbeta - could you have a look at the account dropdown styling?  The avatar is too small.  There's a decent chance that my efforts to fix may be to blame.

Icemanbeta:
@burgwyn All yours!

src/sass/_dropdown.scss
```diff
@@ -20,17 +20,13 @@
 }
 
 .account-dropdown--avatar {
-  border-radius: 20px;
-  height: 20px;
+  // height: 20px;
```
mjoyce91: Still need these comments?

src/Components/Avatar/Avatar.jsx
```diff
@@ -20,9 +19,11 @@ const defaultProps = {
 };
 
 const Avatar = ({ initials, firstName, lastName, className, onClick }) => (
-  <InteractiveElement className={className} onClick={onClick} role="img" aria-label={`${firstName} ${lastName}`}>
-    <div className="tm-avatar">{initials}</div>
-  </InteractiveElement>
+  /* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+  <div className={`tm-avatar ${className}`} onClick={onClick} role="img" aria-label={`${firstName} ${lastName}`}>
```
mjoyce91: @burgwyn - was there a display name we were going to use here for the RC?

src/Components/Avatar/Avatar.jsx
```diff
@@ -0,0 +1,33 @@
...
+  initials: '',
+  firstName: '',
+  lastName: '',
+  className: '',
+  onClick: EMPTY_FUNCTION,
+};
+
+const Avatar = ({ initials, firstName, lastName, className, onClick }) => (
+  /* eslint-disable jsx-a11y/no-noninteractive-element-interactions */
+  <div className={`tm-avatar ${className}`} onClick={onClick} role="img" aria-label={`${firstName} ${lastName}`}>
```
mjoyce91: The other commit I commented on was oudated, so I'll ask here. Do we want to use display_name for the RC?

src/sass/_dropdown.scss
Icemanbeta: thanks, sorry it was late ha. 701f3bd2b1413fadd76d25ce4f3ed6ccaca52473

Icemanbeta:
@burgwyn da7721c11fc4444f4590663ba5b222db030bf3d2 This is passing code climate now!

PR #1640, opened 2018-03-21
Stale Token API Interceptors

Icemanbeta:
- Added an `axios` response interceptor for `axios` to
drop stale tokens when recieving a 401 status code from the API
- Added an `axios` request interceptor for `axios` to  parse and verify token within a `try/catch` to catch injected tokens or invalid tokens, and if valid, `axios` will inject the auth headers before the request fires.
- Updated all files (including sagas mocks) to use an app specific instance of axios from `api.js`

```javascript
// Probably won't need `window`
window.localStorage.setItem('"aa3e0f420717dcf242831dc596d899af20c9d3f4"');
window.localStorage.setItem('"test"');
``` 
via console, would be a pretty accurate test

### DIFF
--- a/src/actions/autocomplete/missionAutocomplete.js
+++ b/src/actions/autocomplete/missionAutocomplete.js
@@ -1,4 +1,4 @@
-import axios, { CancelToken } from 'axios';
+import { CancelToken } from 'axios';
 import api from '../../api';
 
 let cancel;
@@ -9,12 +9,14 @@ export function missionSearchHasErrored(bool) {
     hasErrored: bool,
   };
 }
+
 export function missionSearchIsLoading(bool) {
   return {
     type: 'MISSION_SEARCH_IS_LOADING',
     isLoading: bool,
   };
 }
+
 export function missionSearchSuccess(missions) {
   return {
     type: 'MISSION_SEARCH_FETCH_DATA_SUCCESS',
@@ -27,20 +29,19 @@ export function missionSearchFetchData(query) {
     if (cancel) { cancel(); }
     dispatch(missionSearchHasErrored(false));
     dispatch(missionSearchIsLoading(true));
-    axios.get(`${api}/country/?q=${query}&limit=3`, {
+    api.get(`/country/?q=${query}&limit=3`, {
       cancelToken: new CancelToken((c) => {
         cancel = c;
       }),
-    },
-    )
-      .then((response) => {
-        dispatch(missionSearchIsLoading(false));
-        return response.data.results;
-      })
-      .then(results => dispatch(missionSearchSuccess(results)))
-      .catch(() => {
-        dispatch(missionSearchHasErrored(true));
-        dispatch(missionSearchIsLoading(false));
-      });
+    })
+    .then((response) => {
+      dispatch(missionSearchIsLoading(false));
+      return response.data.results;
+    })
+    .then(results => dispatch(missionSearchSuccess(results)))
+    .catch(() => {
+      dispatch(missionSearchHasErrored(true));
+      dispatch(missionSearchIsLoading(false));
+    });
   };
 }
--- a/src/actions/autocomplete/postAutocomplete.js
+++ b/src/actions/autocomplete/postAutocomplete.js
@@ -1,4 +1,4 @@
-import axios, { CancelToken } from 'axios';
+import { CancelToken } from 'axios';
 import api from '../../api';
 
 let cancel;
@@ -9,12 +9,14 @@ export function postSearchHasErrored(bool) {
     hasErrored: bool,
   };
 }
+
 export function postSearchIsLoading(bool) {
   return {
     type: 'POST_SEARCH_IS_LOADING',
     isLoading: bool,
   };
 }
+
 export function postSearchSuccess(posts) {
   return {
     type: 'POST_SEARCH_FETCH_DATA_SUCCESS',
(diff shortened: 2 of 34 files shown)


src/api.js
```diff
@@ -1,2 +1,40 @@
...
+      /* eslint-disable global-require */
+      require('./store').dispatch(logoutRequest());
+      /* eslint-enable global-require */
+      break;
+
+    default:
+      // TODO:
+      // We don't want to stop the pipeline even if there's a problem with the dispatch
+      // and if there is, that should be resolved. This is just a placeholder until we
+      // actually need a default case to satisy eslint.
```
burgwyn: satify -> satisfy

Icemanbeta: thanks!

src/api.js
```diff
@@ -1,2 +1,40 @@
...
+import { fetchUserToken, hasValidToken } from './utilities';
+import { logoutRequest } from './login/actions';
+
+export const config = {
+  baseURL: process.env.API_URL || 'http://localhost:8000/api/v1',
+};
+
+const api = axios.create(config);
+
+api.interceptors.request.use((request) => {
```
mjoyce91: :+1:

src/api.js
```diff
@@ -1,2 +1,40 @@
-const api = 'http://localhost:8000/api/v1';
+import axios from 'axios';
+import { fetchUserToken, hasValidToken } from './utilities';
+import { logoutRequest } from './login/actions';
+
+export const config = {
+  baseURL: process.env.API_URL || 'http://localhost:8000/api/v1',
```
mjoyce91: @burgwyn - we'll need to update the server script to use this

src/api.js
```diff
@@ -1,2 +1,40 @@
...
+});
+
+api.interceptors.response.use(response => response, (error) => {
+  switch (error.response.status) {
+    case 401:
+      // Due to timing of import store before history is created, importing store here causes
+      // exports of api to be undefined. So this causes an error for `userProfile.js` when
+      // attempting to login. Went with the eslint quick re-enable to get around this.
+      /* eslint-disable global-require */
+      require('./store').dispatch(logoutRequest());
```
mjoyce91: Hmm I don't think this is getting called. When I change my token to an invalid one, I'm still "logged in" to the site. Adding a `console.log` to the `logout` saga isn't triggered.

src/login/sagas.test.js
```diff
@@ -20,8 +20,8 @@ describe('login functions', () => {
         password: 'admin',
       })
 
-      // Start the test. Returns a Promise.
-      .run();
+      // Start the test. Returns a Promise. [silent warnings]
+      .silentRun();
```
mjoyce91: What does `silentRun` do?

src/actions/autocomplete/missionAutocomplete.js
```diff
@@ -1,4 +1,4 @@
-import axios, { CancelToken } from 'axios';
```
mjoyce91: Check out the test coverage here, I think something made us lose some coverage.

src/actions/autocomplete/postAutocomplete.js
```diff
@@ -1,4 +1,4 @@
-import axios, { CancelToken } from 'axios';
```
mjoyce91: Same thing in this file as in missionAutocomplete - check out test coverage.

src/api.js
Icemanbeta: Let me Brexit check I forgot too test after doing require...might be a store Import timjng issue again..

src/login/sagas.test.js
Icemanbeta: Silence warnings with the timeout

mjoyce91: Oh nice! I remember we had tried figuring this out awhile ago, but couldn't

src/actions/autocomplete/missionAutocomplete.js
Icemanbeta: Fixed! eb90f6fcb24f017c3d4b89ae34bfbacfa7596100

src/actions/autocomplete/postAutocomplete.js
Icemanbeta: Fixed eb90f6fcb24f017c3d4b89ae34bfbacfa7596100

src/api.js
```diff
@@ -23,7 +23,7 @@ api.interceptors.response.use(response => response, (error) => {
       // exports of api to be undefined. So this causes an error for `userProfile.js` when
       // attempting to login. Went with the eslint quick re-enable to get around this.
       /* eslint-disable global-require */
-      require('./store').dispatch(logoutRequest());
+      require('./store').store.dispatch(logoutRequest());
```
mjoyce91: Good catch on this. Unfortunately we end up in a loop on the login page if the token is stale. This is because there are some components (namely, the Notifications component, though it's hidden) on the login page that perform ajax, so there are constantly bad requests redirecting you back to the login page. This only happens when the token is stale, not when it doesn't exist. I guess there are a couple options - 1. Make sure any component like that doesn't exist on the login page. 2. Only redirect the user to `/login` if they're not already there.

burgwyn: I’ve encountered that race condition in my testing on the login screen too.  We won’t have the login screen in our release, but I’d still like to put a solution in place.

mjoyce91:
I added two commits. One updates the `Notifications` component to only update if the user is not on the `/login` route. The second adds a check to the logout saga to not redirect the user to `/login` route if they're already navigated there - this prevents any race condition from looping the user continuously to the `/login` page.

mjoyce91:
```== src/api.js (1 issue) ==
13: Assignment to property of function parameter 'request'. [eslint]```

This one was weird because it wasn't flagged in Atom. Easy enough to fix though.

burgwyn:
LGTM

PR #1660, opened 2018-03-22
Add filtering by Bid Cycle

mjoyce91:
Adds filtering by Bid Cycle to the Results page.

The only thing I want to confirm with @afrasier is this:
Is the correct way to search for multiple bid cycle the following: `?is_available_in_bidcycle=1,2`


afrasier:
Yes, that's correct.

PR #1665, opened 2018-03-23
Update language filter

mjoyce91:
Updates the language filter to a standard multi-select. The modularity of multi-select filters made this pretty simple!


mjoyce91:
I think saved searches got borked somewhere along the way. The endpoint `api/v1/searches//` is getting called when I try to save a search.

PR #1673, opened 2018-03-24
Rc1 accessibility

burgwyn:


src/sass/_condensedCard.scss
```diff
@@ -85,6 +85,14 @@ $condensed-card-padding: 9px 20px 9px;
 
   .condensed-card-top-header-left {
     float: left;
+
+    h3 {
+      font-family: unset;
```
mjoyce91: Use an IE-friendly rule https://developer.mozilla.org/en-US/docs/Web/CSS/unset#Browser_compatibility

src/sass/_home.scss
```diff
@@ -120,6 +120,10 @@
   .positions-section-title {
     font-family: Merriweather;
     font-size: 19px;
+    font-weight: unset;
```
mjoyce91: Use an IE-friendly rule https://developer.mozilla.org/en-US/docs/Web/CSS/unset#Browser_compatibility

burgwyn:
Closing for #1679

PR #1677, opened 2018-03-26
Redesign Grade Filter

mjoyce91:
Redesigns the Grade filter to use custom sorting and custom descriptions. ~~This is a WIP because there are some questions around which grade filters we should be displaying. If there are no changes, we will need to include the full description for the "OM" grade.~~

~~This was also branched from https://github.com/18F/State-TalentMAP/pull/1672 so we'll want to merge that first.~~

Only displays `is_available` grades in the Grade filter.


src/sass/_results.scss
```diff
@@ -450,7 +450,7 @@ $results-container-width: 100% - $filter-container-width;
   .accordion-checkbox-grade {
     .tm-checkbox {
       float: left;
-      width: 35%;
+      // width: 35%;
```
burgwyn: Remove comment

src/utilities.js
```diff
@@ -87,6 +87,23 @@ export const propSort = propName => (a, b) => {
...
+
+  // if grade is not in sortingArray, push to bottom of list.
+  const indexOfA = sortingArray.indexOf(A) >= 0 ? sortingArray.indexOf(A) : sortingArray.length;
+  const indexOfB = sortingArray.indexOf(B) >= 0 ? sortingArray.indexOf(B) : sortingArray.length;
+
+  if (indexOfA < indexOfB) {
+    return -1;
+  }
+  if (indexOfA > indexOfB) { return 1; }
+  return 0;
```
burgwyn: Throw a test in for this

src/actions/filters/helpers.js
```diff
@@ -1,3 +1,18 @@
+// Attempt to map the non-numeric grade codes to a full description.
+// If no match is found, return the unmodified code.
+export function getCustomGradeDescription(gradeCode) {
+  if (gradeCode === 'CM') {
+    return 'CM Career Minister (FE-CM)';
```
burgwyn: Let's test these labels.  Also, maybe a `switch` for the numeric values that drop through?

mjoyce91: Would the `switch` be to filter out any values that weren't previously matched and also aren't numeric?

burgwyn:
```
switch(gradeCode) {
  case 'CM':
    return 'CM Career Minister (FE-CM)';
  case: 'MC'
    return 'MC Minister-Counselor (FE-MC)';
  case 'OC':
    return 'OC Couneslor (FE-OC)';
  case 'OM':
    return 'Office Manager (OM)';
  default:
    return gradeCode;
}
```

PR #1635, opened 2018-03-20
Implement Results Card Design

Icemanbeta:
- Created reusable layout components `Row`, `Column` for faster templating
- `ResultsCardDataItem` => `DefinitionList` (`components/DefinitionList`)
- `ResultsCardDataPoint` => `Definition` (`components/DefinitionList/Definition`)
- Depreciated `ResultsCardDataSection`
- Update `Results` mock object to the latest from **State-TalentMAP-API`**
- Created reusable definition wrappers: `DefinitionList`, `Definition`
- Created unit tests for all component use cases
- Added `_lists.scss` and `@import` to scss entry file
- Synced latest design guide wireframe palette (with names) to `variables.scss` and cleaned up duplicate styles by referencing the source of truth.
- Cleaned up `variables.scss` and added section comment blocks so it's easier to read
- Added import of `variables.scss` from `uswds` and their bourbon functions so we can extend/implement the core if we need to
Color
- Updated `@import` paths to be module routed instead of relative  in`styles.scss`
- Added `truncated` prop to `Definition` for truncating text overflow via ellipsis
- Updated `InteractiveElement` to allow for more than `div` and `span` due to `usa-button` unfavorable overriding needs
- Added automatic disabling and active states via classes and attributes on `CompareCheck` and `className` proxy prop
- Implemented new design to `ResultsCard`
- Updated `SystemMessages.js` for easier maintenance (alphabetical)

PR #1678, opened 2018-03-26
[RC1] Implement Results Card Design 

Icemanbeta:
https://github.com/18F/State-TalentMAP/pull/1635


mjoyce91:
@Icemanbeta I think `release-candidate-1` should be the target branch

src/Components/ResultsCard/ResultsCard.jsx
```diff
@@ -1,70 +1,152 @@
...
-          <div className="bid-cycle-container-section">
-            <strong>Date Posted:</strong> {postedDate}
+    <MediaQueryWrapper breakpoint="screenMdMax" widthType="max">
+      {(matches) => {
+        const columns = matches ? [5, 6] : [5, 6];
+        return (
+          <div id={result.id} className="results-card">
+            <Row className="header" fluid>
+              <Column columns={columns[0]}>
+                <h3>{result.title}</h3>
```
mjoyce91: Position title should be Source Sans Pro

src/Components/ResultsCardDataSection/__snapshots__/ResultsCardDataSection.test.jsx.snap
```diff
@@ -1,60 +0,0 @@
-// Jest Snapshot v1, https://goo.gl/fbAQLP
-
-exports[`ResultsCardDataSectionComponent matches snapshot 1`] = `
-<div>
-  <div
-    className="usa-width-five-twelfths data-section-left-container"
```
mjoyce91: What if we use `usa-width-one-half` here and add 20% padding-right to the card container? It would make the card on smaller screen widths not quite as condensed, but still match the design.

mar-26-2018 14-23-26

PR #1679, opened 2018-03-26
RC1 - Accessibility

burgwyn:


src/Components/PositionTitle/__snapshots__/PositionTitle.test.jsx.snap
```diff
@@ -4,6 +4,24 @@ exports[`PositionTitleComponent matches snapshot 1`] = `
 <div
   className="position-details-header-container"
 >
+  <HelmetWrapper
+    defer={true}
+    encodeSpecialCharacters={true}
+  >
+    <title />
+    <meta
+      content="undefined 00003027"
```
mjoyce91: `undefined`?

src/Containers/PageMeta/PageMetaContainer.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+  componentWillMount() {
+    this.getPageTitle();
+  }
+
+  getPageTitle() {
+    const { history } = this.props;
+
+    // listen for changes in history
+    history.listen((historyObject) => {
```
mjoyce91: One thing I noticed is that on initial page load, this function is not called, which means no page title is applied. `history.listen` is only called after a location change. Doing something like this seemed to fix it:

```js
setPageTitle(historyObject) {
    // loop through routes
    routes.forEach((route) => {
      if (matchPath(historyObject.pathname, { path: route.path, exact: false })) {
        // set pageTitle if it exists
        if (route.pageTitle) {
          this.setState({ pageTitle: route.pageTitle });
        }
      }
    });
  }

  getPageTitle() {
    const { history } = this.props;

    // perform once on mount
    this.setPageTitle(history.location);

    // listen for changes in history
    history.listen((historyObject) => {
      this.setPageTitle(historyObject);
    });
  }
```

src/Components/ResultsCondensedCardTop/ResultsCondensedCardTop.jsx
```diff
@@ -21,7 +21,7 @@ const ResultsCondensedCardTop = ({ position, type }) => {
           }
         >
           {useType && <span><FontAwesome name={icon} /> | </span>}
-          {position.title}
+          <h3>{position.title}</h3>
```
mjoyce91: Something going on here

screen shot 2018-03-27 at 9 18 26 am

mjoyce91: I think the new `h3` just needs:

```css
    margin-top: 0;
    display: inline;
```

PR #1681, opened 2018-03-26
Exclude “Transfer”-related Tours of Duty

mjoyce91:
Related to #1680


burgwyn:
PR looks good, just awaiting some data validation before proceeding

burgwyn:
I can confirm that there are 2 transfer values that show up in the on-site dataset.

cydneycappello:
@burgwyn @mjoyce91 Do we know how many positions are tied to the TODs with a Transfer? Robin had said the term "Transfer" was back office speak for 0 R&R. But I wonder if these are positions people would actually want to research or bid on?

burgwyn:
@cydneycappello and @mjoyce91 736/5589 have a Transfer ToD

cydneycappello:
@burgwyn what is 736/5589? Is that a position number?

burgwyn:
@cydneycappello - sorry 736 positions out of the total 5589 positions have a Transfer ToD

cydneycappello:
@burgwyn Thanks for clarifying. I think we may have to revisit this with Chris and Robin. It may need to be in TalentMAP. That's a significant number.

mjoyce91:
No longer needed

PR #1704, opened 2018-04-02
Use effective_date for a position's “Posted” date

mjoyce91:
Use `effective_date` as the property to display for a position's "Posted" date. Based on feedback from @afrasier.


mjoyce91:
Looks like there's a time zone difference that's affecting the snapshot.

PR #1706, opened 2018-04-03
Effective date fix

burgwyn:
- Set runtime environment time zone - NY
- Other clean up


.circleci/config.yml
```diff
@@ -3,6 +3,7 @@ default-docker-node: &default-docker-node
   docker:
     - image: circleci/node:6.11.0
       environment:
+        TZ: "/usr/share/zoneinfo/America/New_York"
```
mjoyce91: :+1:

PR #1710, opened 2018-04-03
Include missing parameter in onKeyDown function

mjoyce91:
We were not passing params for keydowns, so clicking would work, but pressing enter would not.

Also sets focus to the text input when opening the Save New Search dialog.


src/utilities.js
```diff
@@ -271,10 +271,20 @@ export const filterByProps = (keyword, props = [], array = []) => {
...
-  if (element) { element.focus(); }
+// Focus an element on the page based on its ID. Pass an optional, positive timeout number to
+// execute the focus within a timeout.
+export const focusById = (id, timeout) => {
+  let element = document.getElementById(id);
+  if (!timeout) {
+    if (element) { element.focus(); }
+  } else {
+    setTimeout(() => {
+      element = document.getElementById(id);
```
burgwyn: Do we need to re-get this element?

PR #1716, opened 2018-04-03
BidCycles Permission for Admin Only

Icemanbeta:
- Integrate `roles` option into `Menu.js` configuration interface
- Integrate `roles` to user's roles check in `ProfileMenu`


src/Components/ProfileMenu/ProfileMenuCollapsed/ProfileMenuCollapsed.jsx
```diff
@@ -16,12 +17,14 @@ const ProfileMenuCollapsed = ({ expand, isCDO, isGlossaryEditor }) => (
     <NavLinksContainer>
       {
         PROFILE_MENU.map((item) => {
+          const missingRoles = difference(item.roles, roles);
```
mjoyce91: :heavy_minus_sign:

PR #1724, opened 2018-04-09
Final Designs (04-04-2018) - Positions

Icemanbeta:
- Update `Position` tree


burgwyn:
We'll revisit in #1722

Issue #1765, opened 2018-04-13
UI Dev - Re-order Search Dialog / Controls / Pills


mjoyce91:
Re-order these sections on the Results Page as designed:

<img width="877" alt="screen shot 2018-04-13 at 12 59 42 pm" src="https://user-images.githubusercontent.com/11576347/38747976-a050be44-3f1a-11e8-8af8-1848896aea2c.png">

PR #1772, opened 2018-04-16
Sorting Updates

mjoyce91:
- Updates the wording of sorting
- Ensures correct use of sorting queries
- Adds a default sort to the Results page


burgwyn:
Can we add a little more space?
screen shot 2018-04-16 at 10 45 11 am

mjoyce91:
@burgwyn that will be covered in https://github.com/18F/State-TalentMAP/issues/1765

PR #1788, opened 2018-04-16
`ResultsContainer` + `SaveNewSearchContainer` Design Updates

Icemanbeta:
- Updated `ResultsContainer` to **2018-04-16**
- Updated `SaveNewSearchContainer` to **2018-04-16**

### DIFF
--- a/src/Components/SaveNewSearchDialog/SaveNewSearchDialog.jsx
+++ b/src/Components/SaveNewSearchDialog/SaveNewSearchDialog.jsx
@@ -10,14 +10,34 @@ class SaveNewSearchDialog extends Component {
     super(props);
     this.submitNewSavedSearch = this.submitNewSavedSearch.bind(this);
     this.submitUpdatedSavedSearch = this.submitUpdatedSavedSearch.bind(this);
+    this.onSubmit = this.onSubmit.bind(this);
+
+    this.state = {
+      type: 'enter',
+    };
+  }
+
+  onSubmit(e) {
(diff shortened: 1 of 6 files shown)


burgwyn:
apr-16-2018 20-25-48

src/Components/SaveNewSearchDialog/SaveNewSearchDialog.jsx
```diff
@@ -47,6 +47,7 @@ class SaveNewSearchDialog extends Component {
           />
         </FieldSet>
         <div className="saved-search-form-buttons">
+          <button className="usa-button-secondary saved-search-form-secondary-button" onClick={onCancel}>Cancel</button>
```
mjoyce91: I feel like Submit should still be the default action when the user presses "enter" while in the text input.

Icemanbeta: Oh is the enter tied to the next button or something?

Icemanbeta:
@burgwyn Sorry didn't notice it

image

src/Components/SaveNewSearchDialog/SaveNewSearchDialog.jsx
Icemanbeta: @mjoyce91 I made some updates and you should be able to hit enter to submit the form, in edit mode it saves to the same search where as in a new state it saves to the new state: 5cb066dc7d6f8e37354241c4eee724aa587703dd

PR #1795, opened 2018-04-17
Add filters

### LINKED ISSUE #1780: Add "No danger pay" and "No post differential" option to corresponding filters
Need to add a "No danger pay" and "No post differential" option to filters.

### Current
There is no way for users to opt-out of danger pay or post differential. If they select nothing, then they get all. 

screen shot 2018-04-16 at 2 59 52 pm
screen shot 2018-04-16 at 3 00 00 pm
screen shot 2018-04-16 at 3 00 03 pm

### Desired
Users can select "No danger pay" and/or "No post differential" to return results that have 0% for either Danger pay or Post Differential.
screen shot 2018-04-16 at 2 59 40 pm
screen shot 2018-04-16 at 2 59 46 pm

burgwyn:
Closes #1780

### DIFF
--- a/src/reducers/filters/filters.js
+++ b/src/reducers/filters/filters.js
@@ -130,6 +130,7 @@ const items =
           ],
         },
         data: [
+          { id: 0, code: '0', description: 'No post differential' },
           { id: 5, code: '5', description: '5%' },
           { id: 10, code: '10', description: '10%' },
           { id: 15, code: '15', description: '15%' },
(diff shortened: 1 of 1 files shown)


src/reducers/filters/filters.js
```diff
@@ -130,6 +130,7 @@ const items =
           ],
         },
         data: [
+          { id: 0, code: '0', description: 'No post differential' },
```
mjoyce91: I hadn't realized it before, but these cause invalid `id`s. I created https://github.com/18F/State-TalentMAP/pull/1799 to address the issue.

burgwyn: Do we need an issue?

mjoyce91: Sure, I'll make one.

PR #1829, opened 2018-04-20
Check prop equality to determine re-render of Alert components

mjoyce91:
Possibly fixes accessibility issue of alert message being repeatedly read, as I confirmed that this changed prevented the `Alert` component from being re-rendered when there were no changes. I believe this is due to the nested `message` property not passing equality checks by React.

### DIFF
--- a/src/Components/Alert/AlertAlt/AlertAlt.jsx
+++ b/src/Components/Alert/AlertAlt/AlertAlt.jsx
@@ -1,47 +1,55 @@
-import React from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


src/Components/Alert/AlertAlt/AlertAlt.jsx
```diff
@@ -1,47 +1,55 @@
...
+      case 'warning':
+        icon = 'exclamation-triangle';
+        break;
+      case 'error':
+        icon = 'exclamation-triangle';
+        break;
+      case 'success':
+        icon = 'check-circle';
+        break;
+    // default to info
```
burgwyn: 💯

PR #1833, opened 2018-04-20
SAML Auth Hookup with Basic Auth

Icemanbeta:
- Remove `universal-cookie`
- Update to use `localStorage` again
- Update  alot of random stuff


Icemanbeta:
@burgwyn there's some cert unit test errors not sure how you wanted to handle it. I added a file exists check but realized it errors out anyways later

src/login/Components/LoginForm.jsx
```diff
@@ -1,100 +1,83 @@
...
-  // In this case it will be an object with `username` and `password`
-  submit({ username, password }) {
-    this.props.loginRequest({ username, password });
-  }
+import { EMPTY_FUNCTION } from '../../Constants/PropTypes';
+import { initialState } from '../reducer';
 
-  render() {
-    const {
+/* eslint-disable react/prop-types */
```
mjoyce91: Can we remove and add `handleSubmit` to propTypes?

src/login/Components/TokenValidation.jsx
```diff
@@ -2,12 +2,12 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import queryString from 'query-string';
 import Cookies from 'universal-cookie';
-import { connect } from 'react-redux';
 
 import { EMPTY_FUNCTION } from '../../Constants/PropTypes';
 import Alert from '../../Components/Alert/Alert';
 
-import { tokenValidationRequest } from '../actions';
+// import { tokenValidationRequest } from '../actions';
```
mjoyce91: Remove?

src/login/actions.js
```diff
@@ -24,3 +29,32 @@ export const logoutRequest = function logoutRequest() {
...
+      dispatch(() => ({ type: LOGIN_SUCCESS }));
+
+      // set a stringified version of our token to localstorage on our domain
+      localStorage.setItem('token', JSON.stringify(token));
+
+      // get the user's profile data
+      dispatch(userProfileFetchData());
+
+      // redirect them to home
+      // yield put(push('/'));
```
mjoyce91: I think we need this uncommented

src/login/index.jsx
```diff
@@ -1,43 +1,47 @@
...
+Login.propTypes = props$.types;
+Login.defaultProps = props$.defaults;
+
+const mapStateToProps = state => ({
+  login: state.login,
+});
+
+export const mapDispatchToProps = dispatch => ({
+  onSubmit: (form) => {
+    console.log(form);
```
mjoyce91: needed?

src/routes.js
```diff
@@ -10,4 +10,13 @@ const routesArray = [
   { path: '/loginRedirect', componentName: 'LoginRedirect', pageTitle: 'Login Redirect' },
 ];
 
+switch (process.env.NODE_ENV) {
+  case 'development':
+    routesArray.unshift({ path: '/login', componentName: 'Login' });
```
mjoyce91: Why do we `unshift` if this is already in the array? Should we remove it from the default array?

Icemanbeta: Oops, that was definitely from a previous merge, fixed 8d9ba2d26ce4f946c17e7792cc5089ebd19c3e19

src/routes.js
```diff
@@ -10,4 +10,13 @@ const routesArray = [
   { path: '/loginRedirect', componentName: 'LoginRedirect', pageTitle: 'Login Redirect' },
 ];
 
+switch (process.env.NODE_ENV) {
```
mjoyce91: Shouldn't this section be base on `LOGIN_MODE` instead of `NODE_ENV`?

src/routes.js
```diff
@@ -1,13 +1,21 @@
...
   { path: '/details/:id', componentName: 'Position', pageTitle: 'Position details' },
   { path: '/compare/:ids', componentName: 'Compare', pageTitle: 'Compare Positions' },
   { path: '/about', exact: true, componentName: 'About', pageTitle: 'About' },
   { path: '/tokenValidation', componentName: 'Login', pageTitle: 'Token Validation' },
   { path: '/loginRedirect', componentName: 'LoginRedirect', pageTitle: 'Login Redirect' },
 ];
 
+switch (process.env.NODE_ENV) {
+  case 'development':
+    routesArray.unshift({ path: '/login', componentName: 'Login' });
```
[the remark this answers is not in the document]
Icemanbeta: That's true, actually I think I was thinking that before during merge. Leaving it in and letting express override it.

I'll remove the switch.

src/login/sagas.js
```diff
@@ -74,54 +178,30 @@ function* logout() {
...
+    const { loggingIn } = yield race(races);
 
-      // attempt log in
-      yield call(loginFlow, username, password);
+    if (loggingIn) {
+      if (isSAML) {
+        yield call(login, loggingIn.token);
+      } else {
+        const credentials = {
+          username: loggingIn.username,
```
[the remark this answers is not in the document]
Icemanbeta: This should be fixed, try again!

src/Components/Header/MobileNav/MobileNav.jsx
```diff
@@ -35,7 +36,7 @@ const MobileNav = ({ user, logout, showLogin }) => (
                 showLogin ?
                   <Link to="login" id="login-mobile">Login</Link>
                 :
-                  <Link to="login" id="logout-mobile" onClick={logout}>Logout</Link>
+                  <InteractiveElement type="a" id="logout-mobile" onClick={logout}>Logout</InteractiveElement>
```
burgwyn: Should this be `<Link/>` like the desktop version?

mjoyce91: Would it depend on basic vs SAML auth?

src/axiosConfig.js
```diff
@@ -0,0 +1,14 @@
+import axios from 'axios';
```
burgwyn: Delete

src/index.jsx
```diff
@@ -2,11 +2,14 @@
 import 'babel-polyfill';
 import React from 'react';
 import ReactDOM from 'react-dom';
+import setupAxiosInterceptor from './axiosConfig';
```
burgwyn: Remove

src/index.jsx
```diff
@@ -2,11 +2,14 @@
...
 import './sass/styles.scss';
 import App from './Components/App/App';
 
 import '../node_modules/uswds/dist/js/uswds.min';
 
+setupAxiosInterceptor();
```
burgwyn: Remove

src/login/reducer.js
```diff
@@ -15,58 +19,60 @@ const initialState = {
...
       };
 
+      break;
+
     case LOGIN_SUCCESS:
-      return {
-        errors: [],
-        messages: [],
+      state$ = {
         requesting: false,
```
burgwyn: requesting: false,
errors: [],	
messages: [],

src/Components/Header/MobileNav/MobileNav.jsx
Icemanbeta: `DesktopNav` uses `AccountDropdown`, inside `AccountDropdown` it uses the same non-self closing tag though I don't think this was changed in this PR, it was a merge

mjoyce91: Okay. I think in SAML mode we'd want it to be an `<a>` since we need to hit Express. In basic auth mode, logout is handled on the client, and could be a `<Link>`.

src/actions/userProfile.js
```diff
@@ -63,7 +63,7 @@ export function userProfileFetchData(bypass) {
 
     // use api' Promise.all to fetch the profile and permissions, and then combine them
     // into one object
-    axios.all([getUserAccount(), getUserPermissions()])
+    return axios.all([getUserAccount(), getUserPermissions()])
```
burgwyn: Remove `return`

src/api.js
```diff
@@ -19,20 +19,26 @@ api.interceptors.request.use((request) => {
...
 
-    default:
-      // TODO:
+    default: {
       // We don't want to stop the pipeline even if there's a problem with the dispatch
       // and if there is, that should be resolved. This is just a placeholder until we
       // actually need a default case to satisfy eslint.
+      const serverMessage = propOrDefault(error, 'response.data.detail');
+
+      if (serverMessage === 'Invalid token') {
```
burgwyn: I was seeing `Bad token` or something like that

src/login/index.jsx
```diff
@@ -1,134 +1,49 @@
...
-    successful: PropTypes.bool,
-    messages: PropTypes.array,
-    errors: PropTypes.array,
-  }).isRequired,
+  // Just to be safe we don't too much automated injecting that we'll just take the ones that
+  // are configured on the components
+  const options = pick(props, props$.keys);
+  return (!isSignedIn ?
+    <Element {...options} /> :
+    <Redirect to="/" />
```
burgwyn: Does this `Redirect` get loaded?

src/login/reducer.js
```diff
@@ -1,12 +1,16 @@
...
   LOGOUT_REQUESTING,
   LOGOUT_SUCCESS,
+  TOKEN_VALIDATION_REQUESTING,
 } from './constants';
 
-const initialState = {
+export const initialState = {
   requesting: false,
   successful: false,
   messages: [],
```
burgwyn: Should we add `errors: []`?

src/login/sagas.js
```diff
@@ -1,67 +1,165 @@
...
+  } else {
+    yield put(authRequest(true, credentials.username, credentials.password));
   }
-  return api.post(loginUrl, { username, password })
-    .then(response => response.data.token)
-    .catch((error) => { changeErrorMessage(error.message); });
+
+  // try to call to our loginApi() function. Redux Saga will pause
+  // here until we either are successful or receive an error
+  const { response, error } = yield call(loginRequest, token);
```
burgwyn: I think we can move this block into the isSAML else block:

```
const { response, error } = yield call(loginRequest, token);
if (response) {
  token = response.data.token;
}
```

src/login/reducer.js
burgwyn: Nevermind, this was "below the fold"

src/login/index.jsx
```diff
@@ -1,134 +1,49 @@
...
-      login: {
-        requesting,
-        messages,
-        errors,
-      },
-    } = this.props;
+props$.keys = keys(props$.types);
 
-    return (
-      <div className="usa-grid-full login-container content-container padded-main-content">
```
mjoyce91: I'll make sure to clean up the spacing on this page before merging. Think it just needs a different class applied:

screen shot 2018-04-27 at 9 59 19 am

src/login/index.jsx
Icemanbeta: Only if its a signout

src/login/index.jsx
mjoyce91: Fixed

Icemanbeta: Thanks, looks good to me

src/Components/Header/MobileNav/MobileNav.jsx
mjoyce91: I didn't change anything here. I'm not sure how everything else is configured and how that would play in with this. From my understanding of how things are configured:

If we're using basic auth, we want this to be `Link` OR to disable the `/login` route on the server. The cleaner solution would be to use `Link`, because basic auth doesn't require any communication with the server on logout.

If we're using SAML, this should be an `<a>` because we need the request to hit the server.

mjoyce91: Oh, even though this is an `a`, there's no `href` associated with it, so that kind of negates my whole point :smile:

mjoyce91: I think in either case, they should be consistent like @burgwyn mentioned.

src/api.js
burgwyn: I'm going to dismiss my feedback.  I ran a bad request an authenticated endpoint and got "Invalid token"

src/login/sagas.test.js
```diff
@@ -1,72 +1,99 @@
...
+  jest.resetModules();
+};
 
 describe('login functions', () => {
-  it('can log in and set the client', () => {
-    const mockAdapter = new MockAdapter(api);
-    mockAdapter.onPost('/accounts/token/').reply(200,
-      { token: '12345' },
-    );
+  xit('can set the client upon providing a valid token (SAML Auth)', () => {
```
burgwyn: I'm looking to get this test running and then this PR is good to go.  The issue I'm seeing is that the environment variable is not set, so the test is run in `basic` mode and not `saml`.  cc @Icemanbeta @mjoyce91

mjoyce91: I can take another look

src/login/sagas.js
```diff
@@ -187,7 +186,7 @@ function* loginWatcher() {
   const evaluate = true;
   // Check if user entered already logged in or not
   while (evaluate) {
-    const isSAML = auth.isSAMLAuth();
+    const isSAML = (process.env.LOGIN_MODE === 'saml');
```
mjoyce91: On line 64, we simply check for `isSAMLAuth: () => (auth.mode() !== 'basic'),`, but here we're checking explicitly for `'saml'`.

burgwyn: We're defaulting to `basic` I believe

burgwyn: Pushed an update for consistency

src/login/sagas.js
```diff
@@ -192,7 +192,7 @@ function* loginWatcher() {
   const evaluate = true;
   // Check if user entered already logged in or not
   while (evaluate) {
-    const isSAML = (process.env.LOGIN_MODE === 'saml');
+    const isSAML = (process.env.LOGIN_MODE !== 'basic');
```
mjoyce91: I guess I was thinking more along the lines of changing the code near line 64. There would be an inconsistency if, for example, `LOGIN_MODE` was undefined:

```js
mode: () => (process.env.LOGIN_MODE || 'basic'), // would return 'basic'
isBasicAuth: () => (auth.mode() === 'basic'), // would return true
isSAMLAuth: () => (auth.mode() !== 'basic'), // would return false
```

Whereas this code on line 195:
```js
const isSAML = (process.env.LOGIN_MODE !== 'basic'); // would return true
```

So `isSAMLAuth` and `isSAML` would return different values, which would probably break something.

PR #1834, opened 2018-04-23
Add a new script to use for local testing

mjoyce91:
Adds a test script to use for local testing. Differences include:

- Removes `--runInBand` (needed for CircleCI) which cuts down test time in half
- Adds `-u` so that snapshots are always updated


burgwyn:
Can you push this to dev too?

mjoyce91:
Sure, since this is in `release-candidate-1`, it will be included in the migration to `dev`.

PR #1836, opened 2018-04-25
Add pm2 and deployment instructions

mjoyce91:


DEPLOY.md
```diff
@@ -91,10 +91,17 @@ sudo cp -r build/* /var/www/html/ -v
...
-/usr/bin/nohup node server.js &
+Ensure pm2 is installed globally:
+
+```
+npm install pm2 -g
+```
+Then use pm2 to start the node server:
+
+```
+pm2 start /var/www/server.js
```
mjoyce91: Include in deployment script?

We could also make use of cluster mode with `pm2 start server.js -i <processes>`. Then we could use `pm2 reload all` to avoid downtime.

Icemanbeta: Would this be better as dependency and a script to startup? We can add in another configuration object like `prodDependencies`. 

Or we can forget the install script and use [optionalDependencies](https://docs.npmjs.com/files/package.json#optionaldependencies) and add a script command for running pm2 which would require the optional install for the command to work

mjoyce91: I think pm2 needs to be installed globally, but it could be a script

burgwyn:
Closing as we're using a different process manager

PR #1838, opened 2018-04-26
Add bid list button back into condensed cards

mjoyce91:
Adds the Bid List button back into the condensed cards and changes the Favorites button to `hideText`.

### DIFF
--- a/src/sass/_buttons.scss
+++ b/src/sass/_buttons.scss
@@ -135,9 +135,26 @@
 }
 
 .favorite-container {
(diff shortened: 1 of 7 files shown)


src/sass/_buttons.scss
```diff
@@ -135,12 +135,29 @@
 }
 
 .favorite-container {
-  &.usa-button {
+  &.usa-button:not(.usa-button-secondary) {
```
burgwyn: 🚫

PR #1846, opened 2018-04-26
Add Notifications/Inbox icons back in

mjoyce91:


src/sass/_iconAlert.scss
```diff
@@ -5,16 +5,16 @@
```
Icemanbeta: Ever wondered why we keep having to use odd px? lol cause USWD has odd numeric baselines 😞

PR #1847, opened 2018-04-27
🐛 Defending against null location 🐛

### LINKED ISSUE #1845: Defect - PostFilter.jsx
I believe this is related to #1843 

Exception when country is null

```
  getAllDomesticCodes(props = this.props) {
    return props.item.data.slice().filter(b => b.location.country === 'United States');
  }

  getAllOverseasCodes(props = this.props) {
    return props.item.data.slice().filter(b => b.location.country !== 'United States');
  }
```

burgwyn:
Closes #1845

### DIFF
--- a/src/Components/SearchFilters/PostFilter/PostFilter.jsx
+++ b/src/Components/SearchFilters/PostFilter/PostFilter.jsx
@@ -60,11 +60,11 @@ class PostFilter extends Component {
   }
 
   getAllDomesticCodes(props = this.props) {
-    return props.item.data.slice().filter(b => b.location.country === 'United States');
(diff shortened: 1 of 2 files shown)


src/Components/SearchFilters/PostFilter/PostFilter.jsx
```diff
@@ -60,11 +60,11 @@ class PostFilter extends Component {
   }
 
   getAllDomesticCodes(props = this.props) {
-    return props.item.data.slice().filter(b => b.location.country === 'United States');
+    return props.item.data.slice().filter(b => (b.location && b.location.country && b.location.country === 'United States'));
```
mjoyce91: The `&& b.location.country` shouldn't be needed since that's already evaluated at the end, but this works.

PR #1855, opened 2018-04-27
Remove "How to Bid" section, Add "Post Details" button, Add "Add to Bid List" button

mjoyce91:
- Removes the "How to Bid" section
- Moves the "Post Details" button
- Adds the "Add to Bid List" button

Relies on https://github.com/18F/State-TalentMAP/pull/1848

PR #1859, opened 2018-04-30
Results cards - Bid List button and Bid Counts

mjoyce91:
Adds the Bid List button and Bid Counts to the results cards.

Relies on #1855 and #1838.


src/Components/ResultsContainer/ResultsContainer.jsx
```diff
@@ -3,7 +3,8 @@ import PropTypes from 'prop-types';
 import PaginationWrapper from '../PaginationWrapper/PaginationWrapper';
 import ResultsList from '../ResultsList/ResultsList';
 import { POSITION_SEARCH_RESULTS, EMPTY_FUNCTION, SAVED_SEARCH_MESSAGE, SAVED_SEARCH_OBJECT,
```
Icemanbeta: I'm surprised eslint lets you do this, I swear I did this before and it say the tab length was incorrect. I think there's an `/* eslint-disable */ somewhere persisting. No complaints here though lol

src/Components/ResultsContainer/ResultsContainer.test.jsx
```diff
@@ -70,6 +72,8 @@ describe('ResultsContainerComponent', () => {
       newSavedSearchHasErrored={false}
       newSavedSearchIsSaving={false}
       resetSavedSearchAlerts={() => {}}
```
Icemanbeta: This goes back to the eslint thing, I swear I've seen eslint errors for not having a default value on required props.

image
https://github.com/airbnb/javascript/blob/master/packages/eslint-config-airbnb/rules/react.js

But actually, why not just set the default value to an empty function. Seems like it doesn't hurt.

src/sass/_results.scss
```diff
@@ -147,7 +177,7 @@ $results-container-width: 100% - $filter-container-width;
       }
     }
 
-    .usa-button {
```
Icemanbeta: Does this work in IE10?

https://caniuse.com/#search=%3Anot

Is there better browser compatibility documentation somewhere? I feel like I don't trust caniuse. But anyway, according to this it doesn't support selector format but it does if it's an element?

src/Components/ResultsContainer/ResultsContainer.test.jsx
mjoyce91: It's the other way around - un-required props need default values.

src/Components/ResultsContainer/ResultsContainer.jsx
mjoyce91: I've noticed that too

src/sass/_results.scss
mjoyce91: We're just targeting IE11. I don't think that site is right - take a look at https://www.w3schools.com/cssref/sel_not.asp

Also seems to look fine when I tested it in IE11 anyways

PR #1865, opened 2018-05-01
Integrated `highlightPosition` actions with Interface

Icemanbeta:
- Hooked up actions to `PositionDetails` tree
- Added reducers to accompany actions for `highlistPosition`
- Include `is_superuser` computed property in `userProfile` action fetch
- Implemented `ServiceNeededToggle` with `is_superuser` permissions
- Added unit tests

image


src/Components/PositionDetailsItem/ServiceNeededToggle/ServiceNeededToggle.jsx
```diff
@@ -0,0 +1,109 @@
...
+
+    if (state.loading) {
+      wrapper.className.push('ds-c-spinner');
+    }
+
+    wrapper.className = wrapper.className
+      .join(' ')
+      .trim();
+
+    return isSuperuser ?
```
mjoyce91: We could abstract the `isSuperuser` out of the component so that in the future, it could be configured to render based on different permissions.

mjoyce91: I.e., when perform that check when it's actually declared:

```jsx
isSuperUser && <ServiceNeedToggle />
```

```jsx
isEuropeanPostsAdmin && <ServiceNeedToggle />
```

etc...

src/Components/PositionDetailsItem/ServiceNeededToggle/ServiceNeededToggle.test.jsx
```diff
@@ -0,0 +1,35 @@
...
+const userMock = {
+  is_superuser: true,
+};
+
+const props = {
+  position: mock,
+  userProfile: userMock,
+};
+
+describe('ServiceNeededToggle', () => {
```
mjoyce91: Can we bump up code coverage for:

`shouldComponentUpdate`
`onClick`
`if (state.checked) {...`
`Iif (state.loading) {...`

src/reducers/positionDetails/positionDetails.js
```diff
@@ -1,3 +1,5 @@
+import { find, merge } from 'lodash';
+
 export function positionDetailsHasErrored(state = false, action) {
```
mjoyce91: Needs test file

src/utilities.js
```diff
@@ -494,3 +494,13 @@ export const redirectToLogout = () => {
   const prefix = process.env.PUBLIC_URL || '';
   window.location.assign(`${prefix}${LOGOUT_ROUTE}`);
 };
+
+export const difference = (object, base) => transform(object, (result, value, key) => {
```
mjoyce91: Can you add description of purpose of this function?

src/Components/PositionDetailsItem/ServiceNeededToggle/ServiceNeededToggle.jsx
Icemanbeta: Im game for dynamic solutions. I added is_superuser to the userProfile object that gets set in the action.

src/Components/PositionDetailsItem/ServiceNeededToggle/ServiceNeededToggle.test.jsx
Icemanbeta: lol so mean

src/reducers/positionDetails/positionDetails.js
Icemanbeta: True, I'll add one tomorrow 😄

src/utilities.js
Icemanbeta: Shit forgot I did end up using it, but I think I'll leave it in since this might be useful. I add comments to this but actually what this is for is to get a deep diff on 2 objects. I actually tried this on the prop object in PositionDetails, and the performance was unnoticeable and it worked.

Essentially, you'd probably use it in like `shouldComponentUpdate` and pass in `this.props` and the `newProps` passed in by react. It'll return you a new object of only the differences. so if one key from one of the props gets changed, it should only give you like:
```
{
  prop1: {
     is_notification: true,
  }
}
```
Speaking of given the new value or the base value, would you prefer the base object as the first argument and the new object the second or like this?

Even if it's not actually used as the final should render logic, I think it'll be good to have for debugging state changes in redux.

mjoyce91: Oh that's cool! I think either way on the params works. If we're going to keep it, let's add some tests.

src/Components/PositionDetailsItem/ServiceNeededToggle/ServiceNeededToggle.jsx
mjoyce91: I guess a better example would be the position property `description.is_editable_by_user`. We could have the same thing exposed on the position object, but for "is highlight-able by user". I actually think the way we have it right now is fine, it may just be worth adding a comment for that somewhere in the component.

Icemanbeta: Oh on the other side, I guess instead of having permission logic in the component, rather in the data source. So what I have right now is:

```js
export function userProfileFetchData(bypass) {
  return (dispatch) => {
    if (!bypass) {
      dispatch(userProfileIsLoading(true));
      dispatch(userProfileHasErrored(false));
    }

    /**
     * create functions to fetch user's profile and permissions
     */
    // profile
    const getUserAccount = () => api.get('/profile/');
    // permissions
    const getUserPermissions = () => api.get('/permission/user/');

    // use api' Promise.all to fetch the profile and permissions, and then combine them
    // into one object
    axios.all([getUserAccount(), getUserPermissions()])
      .then(axios.spread((acct, perms) => {
        // form the userProfile object
        const account = acct.data;
        const permissions = perms.data;
        const newProfileObject = {
          ...account,
          is_superuser: indexOf(permissions.groups, 'superuser') > -1,
          permission_groups: permissions.groups,
        };

        // then perform dispatches
        dispatch(userProfileFetchDataSuccess(newProfileObject));
        dispatch(userProfileIsLoading(false));
        dispatch(userProfileHasErrored(false));
        dispatch(userProfileFavoritePositionHasErrored(false));
        dispatch(userProfileFavoritePositionIsLoading(false));
      }))
      .catch(() => {
        dispatch(userProfileHasErrored(true));
        dispatch(userProfileIsLoading(false));
        dispatch(userProfileFavoritePositionIsLoading(false));
      });
  };
}
```
Then I pass in the `userProfile` object from the container to this component and it just checked that property.

I guess we'd somehow set the property on the incoming position details object from the api action right...or have that property set by the API.

src/reducers/positionDetails/positionDetails.js
Icemanbeta: 600c3d8d0acb7b85aacf9e9f22b1792789c5f42a

src/Components/PositionDetailsItem/ServiceNeededToggle/ServiceNeededToggle.jsx
```diff
@@ -0,0 +1,86 @@
...
+    };
+
+    const isSuperuser = user.is_superuser;
+    const options = {
+      className: ['usa-button', props.className],
+      onClick: this.onClick,
+    };
+
+    if (this.checked) {
+      options.className.push('usa-button-active');
```
mjoyce91: This seems a little odd to use the `active` class. I feel like the text already communicates the purpose of the button.

Icemanbeta:
@mjoyce91 I think all that's left is for @burgwyn to test it!
@burgwyn I switched the target branch for the PR! It's ready for you to run through!

PR #1873, opened 2018-05-04
🐛 Adding SSO certificate and sample config 🐛

burgwyn:
I was working on the build out of the TST environment and realized this was hardcoded.


burgwyn:
@amirbey - Mike was able to fix the failing test case (it failed about 5% of the time)

PR #1955, opened 2020-09-08
Tm 1950 Prevent Search button from widening

vijaynai19:
To go over this change for the widening Search button


vijaynai19:
Bad Request

